PR #285, opened 2022-08-04
BA-161 - Introduced gallery component

hlodnb:
Created Gallery component according to 
https://inovaciebratislava.atlassian.net/browse/BA-161


strapi/patches/@strapi+plugin-graphql+4.3.4.patch
```diff
@@ -0,0 +1,50 @@
+diff --git a/node_modules/@strapi/plugin-graphql/server/services/builders/entity.js b/node_modules/@strapi/plugin-graphql/server/services/builders/entity.js
```
mpinter, line 1: please remove this from the PR before merging - if it makes sense to have it, we can do that in a separate PR

mpinter (approved):
Seems to work fine locally - the build error is fine and due to changes in strapi in this PR not being available yet on staging.

Would do a prod release with what we have now & then try merging it, testing thoroughly o nstaging and if ifne proceeding with prod

PR #310, opened 2022-09-02
BA-682 Change translations and url in road closures section

JacobMu:
JIRA ticket: https://inovaciebratislava.atlassian.net/browse/BA-682

I changed translations for section tab title and button for more news in the bottom part of the section. Plus, extracted logic for similar two tabs that shared similar design to one component.


Balros (changes requested):

Balros:
Also we recently switched to using strict typescript. We are not switching whole project  to strict in one step but we are switching gradually. Therefore please erase first line(`// @ts-strict-ignore`) of all files you touched and resolve any issues that comes up.

Same thing for eslint errors. Erase all comments which disables eslint rules at the begining of the file and resolve any issues that comes up.

mpinter (approved):
When I'm on large screen I get different message on closures than on smaller ones - I assume this is by accident and the large one is incorrect:

Screenshot 2023-02-02 at 09 36 56

Screenshot 2023-02-02 at 09 37 18

Please also see the comment on exporting functions vs a static object, but otherwise looks ok - look at these two before merging but you can merge without further review.

radoslavzeman:
Outdated. The section will undergo redesign.

PR #319, opened 2022-09-26
Task Form 166 and 167

AntonZezul:


next/components/forms/widgets/ExampleInput.tsx
```diff
@@ -10,6 +13,15 @@ export const ExampleInput = (props: any) => {
...
       />
+      <Button variant='link-black' href='/stackoverflow.com/' label='Link value'></Button>
+      {/* <Button variant='link-brand' size='sm' href='https://stackoverflow.com/' label='Link value'></Button> */}
+      <Button size="sm" icon={<SearchIcon />} variant="brand-outline"></Button>
+      <Button text="Button" variant="black" size="sm"></Button>
+      <Button icon={<SearchIcon />} variant="black-outline"></Button>
+      <Button text="Button" variant="negative"></Button>
+      <Button text="Button" variant="plain-brand"></Button>
+      <Button text="Button" variant="plain-black" size="sm"></Button>
+      <Button text="Button" variant="plain-negative"></Button>
```
mpinter, line 24: I'm all for building a 'showcase' page where you can see all the variants next to each other, but ExampleInput.tsx is not a right place for it

ok to keep as is in this PR, but we should move it afterwards - will create a subsequent task, the goal will be to have something like this: https://marianum-next.staging.bratislava.sk/styleguide

next/package.json
```diff
@@ -18,6 +18,7 @@
     "@azure/msal-node": "^1.12.0",
     "@cloudflare/json-schema-walker": "^0.1.1",
     "@js-joda/core": "5.2.0",
+    "@react-aria/button": "^3.6.1",
```
mpinter, line 21: no need for this if you're adding the whole `react-aria` library - you can import the button hooks directly from there

please remove this package and do so

next/components/forms/widgets/ExampleInput.tsx
```diff
@@ -10,6 +13,15 @@ export const ExampleInput = (props: any) => {
         required={props.required}
         onChange={(event) => props.onChange(event.target.value)}
       />
+      <Button variant='link-black' href='/stackoverflow.com/' label='Link value'></Button>
```
mpinter, line 16: Empty components should be self-closing (eslint should complain about this).

```suggestion
      <Button variant='link-black' href='/stackoverflow.com/' label='Link value' />
```

next/components/forms/Button.tsx
```diff
@@ -0,0 +1,181 @@
+import { AriaButtonProps, useButton } from 'react-aria'
+import { LinkButtonProps } from '@react-types/button'
+import React, { forwardRef, ReactNode, RefObject } from 'react'
+import cx from 'classnames'
+
+import SearchIcon from '../../assets/images/arrow-right.svg'
```
mpinter, line 6: confusing naming

```suggestion
import ArrowRightIcon from '../../assets/images/arrow-right.svg'
```

mpinter (approved):
:+1: please fix the few comments below, but you can merge without further review

mpinter:
@AntonZezul I've not really tested all of the possible variants - in case I've missed anything I'm counting on this task :) https://inovaciebratislava.atlassian.net/browse/FORM-190 - this will also create a better place to put the existing example Buttons to.

PR #322, opened 2022-09-29
FORM-170: Create InputField.tsx, add icons

AntonZezul:


mpinter (approved):
code looks fine, will recheck styling against figma once we place it into the styleguide

:+1: 🚢

PR #326, opened 2022-10-03
FORM-168 radio button

AkihitoVM:


next/tailwind.config.js
```diff
@@ -104,7 +104,10 @@ module.exports = {
           universal: {
             100: 'var(--universal-gray-100)',
             200: 'var(--universal-gray-200)',
+            300: '#C2C2C2',
             500: 'var(--universal-gray-500)',
+            600: '#5C5C5C',
+            700: '#333333',
```
mpinter, line 110: I know some colors are defined here and some outside - and frankly, our styles setup is a mess at the moment

but to start moving it in the right direction:
- let's add any new colors as variables to bratislava.css, so they are (mostly) in one place
- the forms colors may have slightly different color pallete - let's keep them clumped together and keep any new colors added prefixed with 'forms-'
- in the tailwind config as well - keep them clumped together under `forms` subobject (colors -> forms -> gray -> name-of-the-gray-variant)

By the way - the reason colors are defined are as variables is to allow for dynamic switching of color themes in different sections - on some subpages, some of these variables are overridden to provide different pallete. Much to be desired in terms of organisation still, but just to explain this one part :)

next/tailwind.config.js
```diff
@@ -124,6 +127,7 @@ module.exports = {
           'universal-800': 'var(--universal-red-800)',
           'universal-500': 'var(--universal-red-500)',
           'universal-300': 'var(--universal-red-300)',
+          'negative-700': '#EF1919',
```
mpinter, line 130: same as with the gray colors

next/components/forms/Radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+      // disabled
+      'opacity-50 cursor-not-allowed': isDisabled,
+    }
+  )
+
+  const containerStyle = cx('flex', rest.className, {
+    'flex-row items-center p-0 gap-4 w-[320px] h-8 left-[20px] top-[20px]': variant === 'basic' && !error,
+    'flex-row items-center p-0 gap-4 w-56 h-12 py-3 px-4 bg-white border-2 border-solid':
+      variant === 'boxed',
+    'border-[#D6D6D6] rounded-[8px]': variant === 'boxed' && !error,
```
mpinter, line 41: you shouldn't need to use [] as heavily - color should be defined and if 8px rounded step should exist and doesn't, you can add it to config

next/components/forms/Radio.tsx
```diff
@@ -0,0 +1,70 @@
...
+    'border-negative-700 rounded-[8px]': variant === 'boxed' && error,
+
+    'flex-col justify-end items-start p-6 w-[384px] h-32 left-[20px] top-[20px] border-2 border-solid border-[#D6D6D6] rounded-[8px]':
```
mpinter, line 44: the fixed width is suspicious, but if you are sure about it I'll recheck once it's placed in the styleguide

next/components/forms/RadioGroup.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { useRadioGroup } from 'react-aria'
+import { useRadioGroupState } from '@react-stately/radio'
```
mpinter, line 3: while this is probably added with react-aria if it does not complain, would still expect it to show up as implicit dependency in package.json - ? 

if it's not a special case handled like this on purpose please add this package

mpinter (changes requested):
Please, check the comments below, as well as:

- rebase this PR on top of master - afterwards react-aria should be already added and /styleguide page should be available
- please add all the new components into styleguide page (will recheck them afterwards)
- this is towards the other pull-requests open - can point them to be based from each other so that the subsequent ones do not include the previous commits. But even better - if they don't depend on each other you can make them contain only the commits relevant to that PR.

PR #329, opened 2022-10-04
Feat/field header

matejpeluha:
Create components FieldHeader and FieldErrorMessage for fields to unify design and logic.


next/components/forms/FieldErrorMessage.tsx
```diff
@@ -0,0 +1,21 @@
...
+  errorMessageProps?: DOMAttributes<never>
+}
+
+const FieldErrorMessage: FC<FieldErrorMessageProps> = ({errorMessage, errorMessageProps}) => {
+  const errorStyle = cx(
+    'mt-1 text-sm text-error'
+  )
+
+  return errorMessage ? (
+    <div className={errorStyle} {...errorMessageProps}>
```
mpinter, line 15: no need to make this more verbous than it needs to be

```suggestion
  return errorMessage ? (
    <div className="mt-1 text-sm text-error" {...errorMessageProps}>
```

next/components/forms/FieldHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+) => {
+  // STYLES
+  const labelStyle = cx(
+    'relative mb-1 text-default font-semibold text-universal-black',
+    {'after:content-["*"] after:ml-0.5 after:absolute after:-top-0.5 after:text-red-brick after:text-xs': required}
+  )
+
+  const descriptionStyle = cx(
+    'mb-1 text-sm text-universal-black'
+  )
```
mpinter, line 33: same as above, this can be inline

mpinter (changes requested):
Please add to /styleguide page as well.

mpinter (approved):
:+1:

looks great - there will probably be some followups with 'Optional' and tooltips (as you've mentioned) but good to go as is - will merge so that it's available

PR #331, opened 2022-10-04
Form 171/fields select

matejpeluha:
Create a new field component with Select functionality. I used the external component Select from react-select and change his styles and also functionality inside our component. I also added examples of SelectField into /styleguide.


mpinter (changes requested):
Please match the designs in figma exactly - font sizes, icon colors, the look of the dropdown button etc.

Would expect the mouse to be 'pointer' style when over the dropdown caret. Would not expect the X to have highlighted background (also, the X itself doesn't look quite right).

The state when field is marked as both required (with the red asterisk) and optional at the same time is invalid and shouldn't be possible. As discussed on the call with designer - the 'Optional' text isn't related to the tooltip, it's an (optional :) ) way of marking field explicitly as 'Optional'. The tooltip itself - the help icon - on the other hand, makes sense only when we're passing a tooltip text into the field - would consider changing this param to string with the tooltip text itself and showing the icon if this string is non-empty.

Regarding icon-help asset - when adding new assets as parts of forms please add them into /assets/forms subdirectory.

mpinter (changes requested):
There are still differences when compared to figma designs - let's go over it on a call sometime on Monday (i.e. after the standup).

mpinter (approved):
The behaviour of the X button still isnt right - watch the corner on hover.

https://user-images.githubusercontent.com/2327312/194883001-59c0c2e0-32a2-4b0d-93b5-1af6a56ff3c0.mov

The font-size inside the multivalue 'labels', while set in your code to 20px explicitly, is then reduced to 85% by some other piece of react-select css. Ideally if you can find where this is coming from and turn that off as on the screenshot below (you can get to this class if you inspect the text inside these 'labels'):

Screenshot 2022-10-10 at 15 58 36

Otherwise looking good :+1: , try to tackle these but you can merge without further review

PR #335, opened 2022-10-10
WILL BE REMOVED Form-168 Form-178 Form-169 Form 176 Radio Toggle Checkbox Alert

AkihitoVM:


next/components/forms/Alert.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const alertContainer = cx('flex flex-row w-[440px] items-start gap-3 rounded-8 py-4 px-5', className, {
+    'text-[#1F1F1F] flex-col': variant === 'message',
+    'bg-form-alert-error-default-bg text-form-alert-error-default': type === 'error' && !solid,
+    'bg-form-alert-success-default-bg text-form-alert-success-default': type === 'success' && !solid,
+    'bg-form-alert-info-default-bg text-form-alert-info-default': type === 'info' && !solid,
+    'bg-form-alert-warning-default-bg text-form-alert-warning-default': type === 'warning' && !solid,
+
+    'flex-row': variant === 'basic',
+
+    'text-[black]': !solid && variant==='message',
```
mpinter, line 57: again, inline color - also, I do not think a plain 'black' color is used anywhere in Figma designs, but I may be wrong

either way - make sure this matches the figma design exactly and that you use the tailwind [] notation as little as possible (it definitle isn't needed here)

next/components/forms/CheckBoxGroupItem.tsx
```diff
@@ -0,0 +1,126 @@
...
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <path
+            d="M5.49999 9.47504L2.02499 6.00004L0.845825 7.17921L5.49999 11.8334L15.5 1.83337L14.3208 0.654205L5.49999 9.47504Z"
+            fill="white"
+          />
+        </svg>
+      )}
+      {isIndeterminate && (
+        <svg
```
mpinter, line 93: you shoudl be able to set className on svg even if it's defined in it's own file (please corrrect me if I'm wrong) - for such as these, prefer having the svg in separate file and not inline

next/components/forms/SingleCheckBox.tsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react'
```
mpinter, line 1: do we need both singlecheckbox and checkboxgroupitem ?

if there are differences in functionality which are hard to reconcile (and which warrant two separate components) they should at least be backed up by same visual component, since they look the same in the end.

But a completely ideal way would be for SIngleCheckBox and CHeckBoxGroup to work in such way that the former can be placed in the latter and the 'api' works and makes sense.

mpinter (changes requested):
For comments regarding inline colors/svgs/eslint or ts issues please check the whole PR, might not have commented on all of them.

But all in all it's on the right track. The only thing warranting a large rewrite will be the checkbox reconciliation (sorry I couldn't comment on this earlier today, read it just now).

mpinter (approved):
:+1: looks ok - there will probably be some followups, but let's address them afterwards

🚢

PR #336, opened 2022-10-10
Form 179/tag

matejpeluha:
Create component tag with all styling and functionality from figma. When tag is removable, it can not be branded (i understand it like this from figma, but there is no problem to change it) and also if we try to remove tag by clicking on X icon, it will just emit event onRemove.


mpinter (approved):
:+1:

PR #337, opened 2022-10-10
Form 194/style guide brands

matejpeluha:
Refactor style guide because code started to be extremely messy, and with need of use any states to handle events of some future components, this should be much easier to test it and create new showcases of components in styleguide.

I also added buttons for changing brand of style guide, so color branding will change as you wish.


mpinter (approved):
:+1: glad for this cleanup 🧹 

please message the slack channel when you merge this regarding the new styleguide structuring

PR #338, opened 2022-10-10
BA-694 erase hiding scrollbar on webkit devices

Balros:


mpinter (approved):
this indeed shows scrollbar on webkit - though a word of warning, I'm not sure if the core of the BA-694 issue was really about mac scrollbars

PR #339, opened 2022-10-11
FORM-174: Add DatePicker component, add to styleguide page examples

AntonZezul:


AntonZezul:
Problem with TypeScript strict mode

next/components/forms/DatePicker/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,62 @@
...
+          type="button"
+          {...nextButtonPropsBtn}
+        >
+          <ChevronIconRight />
+        </button>
+      </div>
+      <CalendarGrid state={state} />
+      <div className="flex items-center justify-between border-t-2 border-[#333] py-3 px-4">
+        <Button onPress={onClose} text="Zrušiť" variant="plain-black" size="sm" />
+        <Button onPress={onSubmit} text="Potvrdiť" variant="black" size="sm" />
```
mpinter, line 56: we should use i18n libraries for these texts - currently we lack the setup we need for the forms part, so no action required in this PR, just noting there will be a followup

mpinter (changes requested):
:+1: really nice work, just fix the ts issues & move the colors to global config & we're ready to 🚢 

leaving on 'requested changes' as I want to take a quick look at the colors config once it's here, but the second review should go through quickly

mpinter (approved):
✅

PR #342, opened 2022-10-13
Add to FieldErrorMessage.tsx prop disabled

AntonZezul:


mpinter (changes requested):
Alright, didn't know which approach you'd take exactly, but I wouldn't do it this way.

If we want t ochange styling of the error (i.e. display it as gray - but this woudl be up to discussion with the designer) then the disabled prop on error message component makes sense. But if with the prop passed in we don't display anything - as in this PR - I'd much prefer to mvoe this logic one level higher.

That is - if the field that renders error message is disabled, don't render the FieldErrorMessage component at all.

Please change accordingly, in case I'm missing something let me know.

AntonZezul:
Not necessary prop to FieldErrorMessage.tsx

PR #343, opened 2022-10-14
Geolocation for iframe

justMatys:
I have added geolocation switch (for map purposes) which enables geolocation within an iframe.


mpinter (approved):
:+1:

PR #346, opened 2022-10-17
basic style for textarea component

AkihitoVM:


mpinter (approved):

mpinter (changes requested):
Just now realising - we want to be able to interact with this via onChange / value fields, not just ref - please change accordingly.

mpinter (approved):

PR #349, opened 2022-10-17
basic style for alert component

AkihitoVM:


mpinter (changes requested):
- code styling & newlines at the ends of files
- the styles do not match figma exactly (i.e. check font weight on the buttons or padding on bottom of alert without buttons)

mpinter (changes requested):
The font weights are still incorrect - you should always check the styling against figma:

Screenshot 2022-10-21 at 10 31 14

Screenshot 2022-10-21 at 10 31 37

I.e. the button & the title are wrong. I also see that the gaps (or paddings/margins) between sections inside the alert are wrong - we're being a bit more lenient with these as this can get easier together with fixing the tailwind config (will write regarding this to slack). Still, in case you're doing those 'by eye' for them to roughly match the design, please again - look at the exact numbers in figma and try to follow those.

AkihitoVM:
> The font weights are still incorrect - you should always check the styling against figma:
> 
> Screenshot 2022-10-21 at 10 31 14
> 
> Screenshot 2022-10-21 at 10 31 37
> 
> I.e. the button & the title are wrong. I also see that the gaps (or paddings/margins) between sections inside the alert are wrong - we're being a bit more lenient with these as this can get easier together with fixing the tailwind config (will write regarding this to slack). Still, in case you're doing those 'by eye' for them to roughly match the design, please again - look at the exact numbers in figma and try to follow those.

Ako som spominal na calle, typ takeho buttonu nie je implementovany v ramci Buttonov vo figme, je to iba vzorec vyuzitia tlačitka.

next/tailwind.config.js
```diff
@@ -270,6 +300,9 @@ module.exports = {
         61: '244px',
         87: '350px',
       },
+      gap: {
+        3: '13px'
```
mpinter, line 304: still not convinced about this one but let's deal with it in the global spacing rewrite (will create a JIRA task today)

mpinter (approved):
:+1: lgtm

PR #350, opened 2022-10-19
Form-177/upload

matejpeluha:
Create Upload component. I did not use input with type file because its hard to style and not practical. Its possible to use also refs and get "data-value" with it. It wil be needed to finish it in second iteration.

Second iteration: add alerts, add loading bar, add handling of files which will be sent


next/components/forms/Upload/Upload.tsx
```diff
@@ -0,0 +1,122 @@
...
+                          onClick={handleOnClickUpload} />
+          : type === 'dragAndDrop'
+            ? <UploadDropArea ref={ref}
+                              value={value}
+                              sizeLimit={sizeLimit}
+                              supportedFormats={supportedFormats}
+                              disabled={disabled}
+                              isFileBroken={isFileBroken}
+                              onClick={handleOnClickUpload}
+                              onDrop={handleOnDrop} />
```
mpinter, line 106: strange styling - do you have prettier & eslint setup correctly ?

mpinter (approved):
Please look into file (auto)formatting before merging - but otherwise looks good :+1:

PR #351, opened 2022-10-19
feat: add forwardRef to SelectField

matejpeluha:
add forward ref possibility to SelectField

### DIFF
--- a/next/components/styleguide/showcases/SelectFieldShowCase.tsx
+++ b/next/components/styleguide/showcases/SelectFieldShowCase.tsx
@@ -2,7 +2,8 @@ import { Wrapper } from '../Wrapper'
 import { Stack } from '../Stack'
 import SelectField from '../../forms/SelectField'
 import { MultiValue } from 'react-select'
-import { useState } from 'react'
+import React, { useState } from 'react'
+import Select from 'react-select/base'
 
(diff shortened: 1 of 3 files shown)


next/components/styleguide/showcases/SelectFieldShowCase.tsx
```diff
@@ -28,7 +33,12 @@ const SelectFieldShowCase = () => {
         <SelectField name="select-test" label="Select Field" options={selectOptions} disabled/>
       </Stack>
       <Stack>
-        <SelectField name="select-test" label="Select Field" options={selectOptions} required/>
+        <SelectField name="select-test" label="Select Field" options={selectOptions} required ref={ref}
+                     value={selectValueSecond}
+                     onChange={value => {
+                       setSelectValueSecond(value)
+                       console.log(ref.current?.getValue())
```
mpinter, line 40: was here likely just for debug and can be removed

mpinter (approved):

next/components/styleguide/showcases/SelectFieldShowCase.tsx
matejpeluha: Yes this is on purpose, just for showing that ref is working.

PR #352, opened 2022-10-21
FORM-175: Add component TimePicker

AntonZezul:
Add to tailwind.config plugin 'tailwind-scrollbar-hide', change folder name from 'DatePicker' on 'DateTimePicker'.
Need to discuss about a validation, and case when you pick just hour or minute.


mpinter (approved):
would be nice for the component to not show an error while you are still editing (i.e. when you set just the hour and not the minutes yet, you get the error message right away, but really you are still editing and shouldn't get error messages until you are finished) - it's relatively ok for it to work like this for now, but if this is an easy change to be made please do so

considering today's standup talk we may be iterating this further, but current state is fine to be merged 🚢 , just check & address the few comments

PR #354, opened 2022-10-24
Form 185/tooltip

matejpeluha:
Create a tooltip and implement it in FieldHeader. The tooltip position is not absolute by default and also not automatic by default. Software developers can set props absolute, top, bottom, left, and right for positioning the tooltip. You can also add className as you want.


mpinter (approved):
:+1:

PR #355, opened 2022-10-24
FORM-192 pdf transform

kristianduda:


mpinter:
Marking with "needs work" label for now, just so I can easily notice when it's ready 🙂

next/utils/xslt.ts
```diff
@@ -0,0 +1,26 @@
+// @ts-ignore
```
mpinter, line 1: so, this one isn't used and is used only for testing ? or it's something else ?

please add a comment detailing why this file exists, or remove if it's forgotten and shouldn't be here

mpinter (approved):
:+1: 🎉 

good job with the fop setup

did not test this yet, will wait until we can do it 'end-to-end' from the forms ui - code looks good, lgtm

next/utils/xslt.ts
kristianduda: As could not find a declaration file for module 'saxon-js'. 
Commend added.

PR #356, opened 2022-10-25
FORM-192 text and html transform

kristianduda:


mpinter (changes requested):
Otherwise looks good, but I'm afraid I messed it up by removing the validateFormName. Before merging, can you please:

- solve the form name validation
- provide test data to feed to the transformation (assuming you have these)

On the latter note - we'll need to lean into testing more soon as backend functionality is encroaching on this codebase :). So if you setup some jest tests along with this PR it would be only benefical.

PR #359, opened 2022-10-25
FORM-196: Design system - typography

AntonZezul:


AntonZezul:
Sorry for name of PR

next/tailwind.config.js
```diff
@@ -77,16 +77,30 @@ module.exports = {
...
-        xxs: ['12px', '18px'],
-        xs: ['12px', '18px'],
-        sm: ['16px', '24px'],
-        default: ['20px', '30px'],
-        md: ['24px', '36px'],
-        lg: ['32px', '41.6px'],
-        xl: ['36px', '42.3px'],
-        '2xl': ['40px', '52px'],
-        xxxl: ['48px', '62.4px'],
+        'p-xs': ['12px', '18px'],
```
mpinter, line 80: a couple changes proposed with this config - these are the defaults available in tailwind https://tailwindcss.com/docs/font-size#setting-the-font-size , I'd like to prevent someone using one of these options by accident

if you move this config one level higher out of the ''extends" object, it will replace instead of extending. Of couse, this unfortunately requires the hunting down of the default tailwind classes for font sizes but that shouldn't be too bad.

The other suggestion is to align this a bit more with the tailwind approach - that is, use `p-base` and `h-base` instead of p-default and to use 3xl instead of xxxl.

next/pages/_forms-test/[eform].tsx
```diff
@@ -0,0 +1,231 @@
+import { Button, PageHeader, SectionContainer } from '@bratislava/ui-bratislava'
```
Comment (mpinter) [inline next/pages/_forms-test/[eform].tsx:1]: this file was remove recently, please remove as well

next/pages/typography.css
```diff
@@ -0,0 +1,57 @@
...
+    @apply text-p-xs;
+  }
+
+  .text-lbl-1 {
+    @apply text-p-md;
+  }
+
+  .text-lbl2 {
+    @apply text-p-default;
+  }
```
mpinter, line 56: yes, let's rename these as it's not really clear

as they shouldn't be needed that often, what about something really descriptive (and following the tailwind sizes name) ? `text-button-l` and `text-button-base` ?

mpinter (approved):
Looked around the page and it's looking good - would like to take a quick look once you remove the default font sizes in tailwind config as well whether it does not break something horribly :) can do that once deployed to staging as well. 

But otherwise good to go, address the few comments and lgtm 🚢

mpinter (approved):

PR #365, opened 2022-11-03
Form 196/design system typography

Balros:


mpinter (changes requested):
I see multiple mixed `text-` and `font-` classes  (that is, setting both the new class which should set all typography and then overriding the font) - this shouldn't happen, we should always be using one of the available typography styles.

Let's try to force it into those. May require communication with Patrik in places where we can't figure out how to make it work with older components. In edge cases we may create some temporary typography classes, where it will be clear these are just helper to be cleared up together with the components (i.e. `text-temp-carousel`) - but I'd prefer avoiding that completely.

mpinter:
One more thing - if the left-over `font-..` clases are just cases of being left in by accident then al that might be needed is to clear them up :)

mpinter (approved):
:ship:

mpinter (changes requested):
Alright, perhaps one more change before we merge this:

We should be able to easily differentiate in typography.css the classes whcih exist in the new design system from those which we want to get rid of. Also, I'm missing the classes for button fonts (again[ referencing the new design system](https://www.figma.com/file/ctDKMhAPIjLUVNNmq430D4/DS-ESBS%3A-Foundations?node-id=73%3A126&t=AHfNX1u0GKdC9Qai-0)) - if these are explicitly omited because they are already covered in some other text-* class, please don't mind that and add them as well, even if they are the same. This is so that when the button styling would be changed this does not affect unrelated fonts.

next/pages/typography.css
```diff
@@ -3,300 +3,279 @@
...
 
-  .text-btn-lg-bold {
-    @apply font-bold text-btn-lg;
-  }
-  .text-btn-bold {
-    @apply font-bold text-btn-base;
-  }
+  /* Figma template variables - end */
+
+  /* Another variables - start */
```
mpinter, line 87:
```suggestion
  /* Other variables - all of these to be removed as the redesign progresses - start */
```

PR #370, opened 2022-11-04
Form 212/design system spacing

AntonZezul:


next/components/ui/NewsCard/NewsCard.tsx
```diff
@@ -89,7 +89,7 @@ export const NewsCard = ({
 
   return (
     <VerticalCard
-      className={cx(className, 'min-w-[280px] leading-extra-tight')}
+      className={cx(className, 'min-w-66 leading-[1.3]')}
```
mpinter, line 92: TODO continue here check the leading on card component

mpinter (approved):
~lgtm~

mpinter (changes requested):
Previous review submitted by accident, apolgies :) code looks good I just want to check the app running before merging it, which I did not do yet - so a bit more time needed on my side

next/tailwind.config.js
```diff
@@ -237,79 +235,40 @@ module.exports = {
...
+        248: '62rem',
+        256: '64rem',
+        264: '66rem',
+        272: '68rem',
+        280: '70rem', // 1120px
+        288: '72rem',
+        296: '74rem',
+        304: '76rem',
+        312: '78rem',
+        320: '80rem', // 1280px
```
radoslavzeman, line 271: How often are these big values used? I would consider if it is worth to have predefined spacing with values above 100. Isn't it easier and cleaner to wrtie just arbitrary values in brackets for those cases?

AntonZezul: They are never used. But sometimes you need width for example 500-600px and we decide to configure values to 1280px.
I agree maybe it is useless, but for me better if we use values from conf and less use brackets.

next/components/ui/FeaturedBlogs/FeaturedBlogs.tsx
```diff
@@ -17,13 +17,15 @@ export const FeaturedBlogs = ({ blogs }: FeaturedBlogsProps) => {
         </div>
         <div className="flex w-1/2 flex-col gap-y-8">
           {blogs.slice(1, 3).map((blog, index) => (
-            <BlogSearchCard key={index} className="h-52" imageClassName="w-52" item={blog} />
+            <BlogSearchCard key={index} className="h-52" imageClassName="w-[206px]" item={blog.data} />
           ))}
         </div>
       </div>
-      <HorizontalScrollWrapper className="gap-x-4 px-8 lg:hidden">
+      <HorizontalScrollWrapper className="px-7.5 gap-x-4 lg:hidden">
```
mpinter, line 24: 7.5 does not exist (anymore), would result in zero padding

```suggestion
      <HorizontalScrollWrapper className="px-8 gap-x-4 lg:hidden">
```

mpinter (approved):

PR #389, opened 2022-11-14
Form-184/Progress Bar

AkihitoVM:
base styling, need to fix tailwind styles + added responsiveness

### DIFF
--- a/next/components/forms/ProgressBar.tsx
+++ b/next/components/forms/ProgressBar.tsx
@@ -1,0 +1,35 @@
+import cx from 'classnames'
+import {useProgressBar} from 'react-aria'
+
+type ProgressBarBase = {
+  type?: 'success' | 'default'
+  value: number
+  minValue?: number
+  maxValue?: number
+  className?: string
+}
+
+const ProgressBar = ({ type = 'default', value = 0, minValue = 0, maxValue = 100, className }:ProgressBarBase) => {
+  const {
+    progressBarProps
+  } = useProgressBar({value,minValue,maxValue});
+
+  const percentage = (value - minValue) / (maxValue - minValue);
+  const barWidth = `${Math.round(percentage * 100)}%`;
+
+  const progressBarStyleContainer = cx('flex flex-row items-center p-0 gap-4 w-full max-w-[500px] h-6', className, {
+  })
+  return (
+    <div {...progressBarProps} className={progressBarStyleContainer}>
+      <div className={cx('flex flex-column items-center w-full max-w-[440px] h-2 bg-form-input-default rounded-full')}>
+        <div style={{ width: barWidth, height: 10 }} className={cx('rounded-full',{
+          'bg-form-black-default': type === 'default',
+          'bg-form-alert-success-default': type === 'success'
+        })} />
+      </div>
+      <div className={cx('font-normal text-base leading-6 not-italic')}>{value}%</div>
+    </div>
+  )
+}
+
+export default ProgressBar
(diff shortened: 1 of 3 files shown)


mpinter:
Added second label as well to indicate there's more work before it's ready.

next/components/forms/ProgressBar.tsx
```diff
@@ -0,0 +1,35 @@
+import cx from 'classnames'
+import {useProgressBar} from 'react-aria'
```
mpinter, line 2: I see this isn't formated according to eslint rules - please fix your eslint setup and make sure you have 'format on save' enabled

next/components/forms/ProgressBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const ProgressBar = ({ type = 'default', value = 0, minValue = 0, maxValue = 100, className }:ProgressBarBase) => {
+  const {
+    progressBarProps
+  } = useProgressBar({value,minValue,maxValue});
+
+  const percentage = (value - minValue) / (maxValue - minValue);
+  const barWidth = `${Math.round(percentage * 100)}%`;
+
+  const progressBarStyleContainer = cx('flex flex-row items-center p-0 gap-4 w-full max-w-[500px] h-6', className, {
```
mpinter, line 20: I don't think max-w should be hardcoded inside the component - we want the 'user' of the component to determine the width of this, i.e. by controlling the enclosing component

That is, I think you can just remove the max-widths from this and it should be reasonable ? though you may need to do more with the inner div below to make it behave as expected

next/components/forms/ProgressBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  })
+  return (
+    <div {...progressBarProps} className={progressBarStyleContainer}>
+      <div className={cx('flex flex-column items-center w-full max-w-[440px] h-2 bg-form-input-default rounded-full')}>
+        <div style={{ width: barWidth, height: 10 }} className={cx('rounded-full',{
+          'bg-form-black-default': type === 'default',
+          'bg-form-alert-success-default': type === 'success'
+        })} />
+      </div>
+      <div className={cx('font-normal text-base leading-6 not-italic')}>{value}%</div>
```
mpinter, line 30: font PR hopefully merged soon :crossed_fingers: - if so, just note you'll need to revisit the styles on this line

next/components/forms/ProgressBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const percentage = (value - minValue) / (maxValue - minValue);
+  const barWidth = `${Math.round(percentage * 100)}%`;
+
+  const progressBarStyleContainer = cx('flex flex-row items-center p-0 gap-4 w-full max-w-[500px] h-6', className, {
+  })
+  return (
+    <div {...progressBarProps} className={progressBarStyleContainer}>
+      <div className={cx('flex flex-column items-center w-full max-w-[440px] h-2 bg-form-input-default rounded-full')}>
+        <div style={{ width: barWidth, height: 10 }} className={cx('rounded-full',{
```
mpinter, line 25: maybe height should rather be controlled via tailwind class

mpinter (changes requested):

next/components/forms/ProgressBar.tsx
AkihitoVM: 'yarn lint' doesn't give me any errors or warnings for this file, my IDE already have enabled eslint rules

PR #396, opened 2022-11-15
BA-488: Create unused logic for intersection check

JacobMu:
This is a new component + hook w/ logic using `IntersectionObserver` API to make it easier when to show sticky navigation menu and when regular navigation.
It is unused currently, as this is a first of other PRs to refactor entire logic for showing navigation menu.

### DIFF
--- a/next/components/ui/VisibilityIntersection/VisibilityIntersectionWrapper.tsx
+++ b/next/components/ui/VisibilityIntersection/VisibilityIntersectionWrapper.tsx
@@ -1,0 +1,20 @@
+import { FC, ReactNode } from 'react'
+
+import { useIntersectionObserver } from './useIntersectionObserver'
+
+interface Props {
+  children: ReactNode
+
+  onHandleVisibility(): void
+}
+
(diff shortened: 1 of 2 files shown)


next/components/ui/VisibilityIntersection/VisibilityIntersectionWrapper.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { useIntersectionObserver } from './useIntersectionObserver'
+
+interface Props {
+  children: ReactNode
+
+  onHandleVisibility(): void
+}
+
+export const VisibilityIntersectionWrapper: FC<Props> = ({ children, onHandleVisibility }) => {
```
radoslavzeman, line 11: We usually write components like this:
```
const VisibilityIntersectionWrapper = ({ children, onHandleVisibility }: Props) => {
...
}

export default VisibilityIntersectionWrapper
```

But if there is a reason to use `FC<Props>` style than okay :)

radoslavzeman (approved):
LGTM :) 👍

PR #398, opened 2022-11-16
Adds beta version for dev 

skifahrer:
This PR adds beta version of bratiska for dev deployment.


mpinter (approved):
Can't really test as I'm on a borrowed PC without VPN, but hey, seems harmless enough, yolo ?

PR #401, opened 2022-11-22
FORM-166

AntonZezul:
Fix Button after review (merging when all design system tasks will be merged)


AntonZezul:
Question: I don't know what text-color need to write in category and category-outline variants

next/components/forms/Button.tsx
```diff
@@ -188,31 +218,53 @@ const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, PolymorphicProp
...
+      <button
+        type="button"
+        ref={ref as RefObject<HTMLButtonElement>}
+        className={style}
+        {...buttonProps}
+      >
         <div className="flex items-center">
           {startIcon && (
-            <i className={`${size === 'lg' && 'mr-3 h-5 w-5'} ${size === 'sm' && 'mr-2.5 h-4 w-4'}`}>{startIcon}</i>
+            <i
```
radoslavzeman, line 241: Why do we use `<i>` element? It stands for "idiomatic text", not for "icon" :) 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

next/components/forms/Button.tsx
```diff
@@ -188,31 +218,53 @@ const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, PolymorphicProp
...
+              className={`${size === 'lg' ? 'mr-3 h-6 w-6' : ''} ${
+                size === 'sm' ? 'mr-2.5 h-5 w-5' : ''
```
radoslavzeman, line 243: I would prefer to use `cx` instead of ternary operators.

mpinter (approved):
🚢 

we agreed that we will solve text colors in the future, if it's still needed

the only actionable before merging should be removing the 'brand' variant, as it should be covered by 'category'

PR #402, opened 2022-11-22
Form-204/rjsf upload

matejpeluha:


next/components/forms/widgets/UploadWidgetRJSF.tsx
```diff
@@ -0,0 +1,108 @@
...
+      files.push(minioFile)
+    } else if (schema.type === "array" && typeof value !== 'string') {
+      value.forEach(file => {
+        const minioFile = fileNameToMinioFile(file)
+        files.push(minioFile)
+      })
+    }
+
+    return files
+  }
```
mpinter, line 66: no need for the function, just calculate files on each rerender. You can also simplify it quite a bit, something like (did not run this, may need recheck):

```suggestion
    const valueArray = schema.type === "array"  && Array.isArray(value) ? value : value ? [value] : [] 
    const files: UploadMinioFile[] = valueArray.map(fileNameToMinioFile)
```

afterwards use files directly instead of the function you've used previously

next/components/forms/widgets/UploadWidgetRJSF.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const [innerValue, setInnerValue] = useState<UploadMinioFile[]>([])
+
+  useEffect(() => {
+    // I need save multiple info about file, which are not available in possible value of rjsf, but needed DURING upload
+    // I am saving these info only in innerValue of widget
+    // but when I go to previous step of stepper, component are rebuild and I still need (now just) fileName, so I read fileNames and transform it
+    setInnerValue(transformInitialValue)
+  }, [])
```
mpinter, line 75: I like the practice of useEffect always having 'exhaustive-dependencies' - not using this array explicitly for any logic, instead using it as a sanity check that we're running side effects on all observed variables.

If you (like here) want ot be explicit about 'run this once when this is mounted', you can use https://usehooks-ts.com/react-hook/use-effect-once from the library we already include.

TLDR: please change this useEffect to useEffectOnce from the lib

mpinter (approved):
please check out the changes below, after addressing them you can merge without further review

PR #405, opened 2022-11-23
Form-208/rjsf-textarea

matejpeluha:


next/components/forms/TextAreaField.tsx
```diff
@@ -26,18 +26,23 @@ const TextAreaField = ({
...
   className,
   ...rest
 }: TextAreaBase) => {
-  const [valueState, setValueState] = useState<string>(value)
+  const [valueState, setValueState] = useState<string>('')
   const ref = React.useRef<HTMLTextAreaElement>(null)
+
+  const getValue = () => {
+    return rest.onChange ? value : valueState
+  }
```
mpinter, line 39: I don't see the reason for putting this into function (?), instead you can do something like

```
const displayValue = rest.onChange ? value : valueState
```

and then use `displayValue`

mpinter (approved):
Please see the change below before merging, other than that 🚢

PR #413, opened 2022-11-25
Form-187/Modals

ValeriiMedvezhonkov:
Refixed message modal + main modal component, changed styling according to new tailwind config + added new 'success' option for message modal. In styleguide was changed views for all of this modals to be openable/closable


mpinter (approved):
Please fix the two comments before merging (a11y errors & empty onClick), otherwise good to go ✅

PR #418, opened 2022-11-29
DEPRECATED Form-189/FieldGroups

ValeriiMedvezhonkov:
changed input field + textarea field max-w-xs to have max width of parent container, added all variants for Form-189 (Field groups), fix eslint for InputField.tsx


next/components/forms/FieldGroups.tsx
```diff
@@ -0,0 +1,170 @@
+import { UploadMinioFile } from '@backend/dtos/minio/upload-minio-file.dto'
```
mpinter, line 1: please split these into multiple files - perhaps we can create a subfolder, so it will look like `compontents/forms/groups/DateFromTo.tsx`, `compontents/forms/groups/TimeFromTo.tsx` etc.

AkihitoVM:
I will fixing it anyway when i got RJSF task

next/components/forms/Groups/TextareaUpload.tsx
```diff
@@ -0,0 +1,31 @@
...
+  UploadLabel: string
+  TextareaLabel: string
+}
+
+export const TextareaUploadGroup = ({ UploadLabel, TextareaLabel }: TextareaUploadBase) => {
+  const [files6, setFiles6] = useState<UploadMinioFile[]>([])
+  return (
+    <div className={cx('gap-6 w-full flex flex-col items-start p-6')}>
+      <TextAreaField label={TextareaLabel} placeholder="" className="h-[150px]" />
+      <div>alebo</div>
```
mpinter, line 18: texts should come from outside and shouldn't be hardcoded

next/components/forms/Groups/DateFromTo.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type DatePickerBase = {
+  label: string
+  errorMessage?: string
+}
+
+export const DateFromTo = ({ label, errorMessage }: DatePickerBase) => {
+  return (
+    <div className={cx('flex-col flex items-start p-0 gap-4')}>
+      <div className="flex flex-row items-center p-0 gap-6">
```
mpinter, line 13: items shouldn't be centered by default, but left aligned instead. This is true for all the groups

next/components/forms/Groups/InputSelectGroup.tsx
```diff
@@ -0,0 +1,148 @@
+import { EnumOptionsType } from '@rjsf/utils'
```
mpinter, line 1: you'll likely want a way of removing rows in this component as well - check around the team, it's possible someone's been looking into this already, and contact Patrik in the channel for how it should look if it does not exist yet

mpinter (changes requested):

mpinter (changes requested):
Just minor changes - I'm mostly curious about the props and the (im)possibility importing them from the child components - if they were done as they are (mostly without imports) on purpose or because of issues that popped up when trying otherwise let me know. Otherwise let's see if we can clean that up some more.

But other than that looking good.

I'll create one more followup to tackle after or during the RJSF - for field groups like Date from-to, it's natural to validate that the first date comes before the latter one. I'm not sure if we want to do this directly in these components or on the RJSF layer, would leave up to you, just putting it out there.

ValeriiMedvezhonkov:
I'm doing RJSF task for field-group components and I found how can I change some piece of code to be OK with RJSF, but for this task it looks fine

next/components/forms/groups/InputSelectGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <PlusCircleIcon fill="red" />
+              </div>
+            ) : (
+              // 0 IQ solution, I will fix it during other fixes
+              <div className="w-6 h-6" />
+            )}
+          </div>
+        )
+      })}
+      <div className="py-0 pr-0 pl-8">
```
mpinter, line 150: once again py-0 pr-0 does not seem to do anything - please recheck the whole code for any like these

next/components/forms/groups/InputSelectGroup.tsx
```diff
@@ -0,0 +1,163 @@
...
+            {fieldGroups.length > 1 ? (
+              <div
+                className="flex w-6 h-6 justify-center items-center cursor-pointer mt-auto mb-5 rotate-45"
+                onClick={() => removeField(element.id)}
+              >
+                <PlusCircleIcon fill="red" />
+              </div>
+            ) : (
+              // 0 IQ solution, I will fix it during other fixes
+              <div className="w-6 h-6" />
```
mpinter, line 145: while noticing this - I don't really mind the select field getting slightly wider without the button, would perhaps prefer it to an empty div - so let's just remove this and revisit if we ever get complains about the width changes

mpinter (changes requested):
The calendar icon from previous comment was resolved in date-time combination but not in date-date combination. Moreover, the visual issue below is present in both:

Screenshot 2022-12-28 at 14 17 03

ValeriiMedvezhonkov:
@mpinter Probably we should start discuss about mobile version of date[from-to] and time[from-to] group. it will took much less time than rewrite whole render component logic for DatePicker and TimePicker, let me know pls what we want

mpinter:
As this is marked as DEPRECATED, closing and assuming the other one hosts all of it's functionality. If I'm wrong we can always reopen.

PR #419, opened 2022-11-29
FORM-175: Fix Time Picker after review

AntonZezul:


mpinter (changes requested):
Does not work well with touch controls, keeps jumping upon selecting hour and I can't scroll up past the selected value:

https://user-images.githubusercontent.com/2327312/205263658-c5fc7063-ce0a-4240-9873-59cbdea5d3a9.mov

Neither works well with gestures on touchpad as it is way too sensitive. And it really does not make much sense to optimise this for mouse scroll, as when you have mouse you can probably pick time easier by writing the value in via keyboard.

Again, I really strongly advise against controlling the scroll directly, this is super finicky and I'd dare to profess it would take all of your time and we'd still keep revisiting it to fix new bugs on new browsers.

Can we perhaps revert to what is currently in master, and do just a single change - adding something like [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) on the highlighted item in the scroll list everytime it's needed, so that the highlighted option is kept visible ? I'd rather have that than extra 300 lines of code that is hard to maintain.

Optionally, we can push this into backlog and focus on other components first, then return to this. Let's add some time limit to this - if the scrollIntoView approach goes nowhere in another 2 hours let's abandon and return to this later.

mpinter (approved):
:+1: good job on the picker behaviour

one more thing that would be nice to change - if I click outside of the picker while it is open it removes the value. Would probably be best if the kept the _previous_ value (stored somewhere in internal state), but if we don't have that even if it worked the same way as clicking on 'Submit' could be fine - definitely less confusing then outright removing the value filled in.

https://user-images.githubusercontent.com/2327312/205851587-44ed8b59-b811-4d59-aa2c-92f938314893.mov

Please look into the above, but you can merge without further review.

PR #426, opened 2022-12-01
FORM-217: xml to json

kristianduda:


mpinter (changes requested):
Please add tests - such that they take json (fully or partially filled , with or without errors), transform them into xml, transform them back and check whether we get the same json aswe started with

mpinter (approved):

PR #428, opened 2022-12-05
fix colors on search page results, more minor fixes

radoslavzeman:
- tmp fix colors in page results
- remove `minKeywordLength` because it's confusing with debaounced search
- removed header from search page to reduce "void space", new h1 is the search field heading
- some cleanup in changed components

### DIFF
--- a/next/components/ui/PageCard/PageCard.tsx
+++ b/next/components/ui/PageCard/PageCard.tsx
@@ -1,18 +1,18 @@
+import ChevronRight from '@assets/images/chevron-right-large.svg'
+import PageBlueIcon from '@assets/images/page-blue-icon.svg'
+import PageBlueIconSmall from '@assets/images/page-blue-icon-small.svg'
+import PageBrownIcon from '@assets/images/page-brown-icon.svg'
+import PageBrownIconSmall from '@assets/images/page-brown-icon-small.svg'
+import PageGreenIcon from '@assets/images/page-green-icon.svg'
+import PageGreenIconSmall from '@assets/images/page-green-icon-small.svg'
+import PagePurpleIcon from '@assets/images/page-purple-icon.svg'
+import PagePurpleIconSmall from '@assets/images/page-purple-icon-small.svg'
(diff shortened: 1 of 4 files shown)


next/components/ui/PageCard/PageCard.tsx
```diff
@@ -49,12 +49,42 @@ export interface PageCardProps {
...
+  ]
+
   const backgroundColor =
     colors.find((c) => {
       return c.pageColor === pageColor
     })?.backgroundColor ?? 'rgb(var(--color-red--light))'
   const { default: PageIcon, small: SmallPageIcon } = findIconByColor(pageColor)
 
+  console.log('pageColor', pageColor, backgroundColor)
+
```
mpinter, line 87:
```suggestion
```

next/pages/vyhladavanie.tsx
```diff
@@ -62,9 +54,10 @@ const Search = ({ footer, mainMenu, page }: AsyncServerProps<typeof getServerSid
   const [searchQuery, setSearchQuery] = useState<string>(debouncedSearchInputValue)
 
   useEffect(() => {
-    if (debouncedSearchInputValue.length > minKeywordLength) {
+    if (debouncedSearchInputValue !== searchQuery) {
       setSearchQuery(debouncedSearchInputValue)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [debouncedSearchInputValue])
```
mpinter, line 61: are you sure (here and on the other place as well) - it seems to me you'd want this to change with searchQuery as well, otherwise you may be comparing against incorrect value

mpinter (changes requested):
Check the useEffect and let me know please, otherwise fine.

next/pages/vyhladavanie.tsx
radoslavzeman: I think there should not be a problem with it for now, because the change should happen alway in one way from debounced value to search value. 
But I'll examine it properly, I would like to merge this PR and make changes in further PR if needed.

PR #435, opened 2022-12-06
BA-725: add homepage search with suggestions

radoslavzeman:
[BA-725](https://inovaciebratislava.atlassian.net/browse/BA-725)
- add search field with suggestions into homepage hero section
- remove homepage header `title` and `subtitle` and add `headline` (multiline text)
- in vzn search result cards - show document titles instead of total count - this is part of UX improvements with vzns
- white "overlay" (fade) - it was inspired by https://reykjavik.is/en but it is not really an overlay with some added functionality so after discussing with PM, we fade just the image, and just on desktop, because that is what was disturbing; everything what is below the field is not faded (even image on smaller screens)

Notes
- you may need to fill out `headline` in homepage header

TODO

- a lot of parts related to search is duplicated in several components, some parts may be redundant - this needs more love and time to refactor, but at least, now everything should be reliable and readable

Screenshots
image
image
image
image


JacobMu (approved):
lgtm from code perspective

mpinter:
Merging, the failed build is due to strapi changes and therefore ok.

PR #437, opened 2022-12-06
FORM-207: RJSF Date and Time Picker

AntonZezul:


AntonZezul:
In DatePicker used 3 times
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
because of value of useDatePickerState has just type DateValue but i need null when i need to reset a value and have a default value.

mpinter (approved):
We should keep in mind we'll need custom error messages later (perhaps to indicate the proper format), but fine for now, lgtm

PR #438, opened 2022-12-06
Meilisearch bugfix

skifahrer:
This PR adds more time for meilisearch to initialize and properly start. Therefore strapi isnot contacting meilisearch when not ready.


mpinter (approved):
:+1:

PR #439, opened 2022-12-06
Fix: Make cards on homepage clickable properly

radoslavzeman:
Few accessibility and UX quick fixes:
- make all "cards" on homepage clickable properly
  - use `after:absolute after:inset-0` on links in card, and `relative` on card top div, so the link `::after` covers the whole card, therefore the whole card is clickable as a link
- remove cursor-pointer where it is redundant or deceptive
- add `@tailwindcss/line-clamp` plugin and remove custom line-clamp non-reliable css classes
- few more minor visual and UX fixes

I tried to touch as few chunks of code as possible :), just repaired what I aimed for.

### DIFF
--- a/next/components/ui/BANavBar/BANavBar.tsx
+++ b/next/components/ui/BANavBar/BANavBar.tsx
@@ -67,7 +67,7 @@ export const BANavBar = ({
             className="group"
             url="/"
             title={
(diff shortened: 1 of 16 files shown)


mpinter (approved):
🚢

PR #442, opened 2022-12-08
FORM-212: Form spacing

AntonZezul:


mpinter (changes requested):
this is slightly confusing, as the default value (the one you get when you don't input anything) is actually '0' - which is then called 'none' and there's a different 'default'

also the idea is to input this only when needed - so if I don't input anything I shoud default to default. More over, this default should look good enough for most of our usecases without further tinkering - that is, if it seems our common use case is '32px above and 0 below' then we may need different defaults for the top and bottom paddings.

Finally:
- I find it more natural to use margin instead of padding here, seems more in line semantically
- in the 'default' mentioned above, it would be more common to have 32px margin BELOW and 0 above - that is component starting as high as it can and leaving space behind it if-need-be
- why style instead of classes ? then with classes, you wouldn't need this helper or this file at all, you could just use `cx` and it's ability to switch on object keys (please change this also on 'sizeHandler' of input, did not notice it beforehand)

mpinter (approved):

PR #446, opened 2022-12-09
FORM-200: async validation

kristianduda:


next/backend/utils/forms.ts
```diff
@@ -209,19 +171,42 @@ export const removeNeedlessXmlTransformArraysRecursive = (
...
   ajv.addFormat('ciselnik', () => true)
 
   ajv.addKeyword('example')
   ajv.addKeyword('enumNames')
+  ajv.addKeyword('isToken')
+  ajv.addKeyword('disabled')
+  ajv.addKeyword('tooltip')
+  ajv.addKeyword('error')
+  ajv.addKeyword({
+    keyword: 'isPhone',
```
mpinter, line 194: for this not to confuse, as we will not use that endpoint for real phone otp and I can see this being strange to someone opening the project later down the line without context 

Let's rename everything used in this example async validation so that it is clear it's just an example - ie. keyword `isExampleAsyncValidation`, the validate function `exampleAsyncValidation`, endpoint etc etc. Where warranted (i.e. the api endpoint) add additional comments

make it so that if the field validated this way is empty the validation fails, otherwise it passes

furthermore, quite a tricky part of this setup is the question of calling the same validation from FE and from BE - it likely won't be the exactly same call (as FE will need a network call and BE won't), but it shouldn't be two completely different functions either. Find some nice way of how to present and organise this.

next/pages/api/user/verification/generate.ts
```diff
@@ -0,0 +1,21 @@
+import { withSentry } from '@sentry/nextjs'
```
mpinter, line 1: when we change this to use just a single example endpoint you can likely get rid of this generate one completely

next/utils/utils.ts
```diff
@@ -51,4 +52,39 @@ const isServer = () => typeof window === 'undefined'
...
+export const isProductionDeployment = () => process.env.NEXT_PUBLIC_IS_STAGING !== 'true'
+
+export type JsonSchema = JSONSchema7Definition
+interface JsonSchemaProperties {
+  [key: string]: JSONSchema7Definition
+}
+
+export const getAllPossibleJsonSchemaProperties = (
+  jsonSchema: JsonSchema | undefined,
+): JsonSchemaProperties => {
```
mpinter, line 64: probably better suited for forms helper

mpinter (changes requested):
Please rewrite and anotate so that it's clearly just an example (apologies for confusion with the phone validation as sample path, that was meant more to illustrate the issue). Also find a way to reuse / group the BE and FE validate function.

Besides this:
- I see the formatting is slightly off, please make sure your formatter is set up correctly (it respects the eslint/prettier setup in the repo) - ask around in the team if help is needed
- this has been open for some time now (again, apologies), make sure that only the changes related to the task are in the PR (minor refactors done alongside are ok, just check whether changes done against older version of master aren't present)

next/pages/form/[eform].tsx
```diff
@@ -68,15 +68,23 @@ const FormTestPage = ({
   const menuItems = mainMenu ? parseMainMenu(mainMenu) : []
   const router = useRouter()
 
+  let escapedSlug = ''
   const formSlug = forceString(router.query.eform)
-  const pageSlug = `form/${formSlug}`
+  // eslint-disable-next-line unicorn/prefer-regexp-test
+  if (formSlug.match(/^[\da-z-]+$/)) {
```
Comment (mpinter) [inline next/pages/form/[eform].tsx:74]: why not comply to eslint ? if rewriting to using the suggested way brings other errors please add comment detailing

next/jest.config.js
```diff
@@ -12,6 +12,10 @@ const customJestConfig = {
   setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
   // if using TypeScript with a baseUrl set to the root directory then you need the below for alias' to work
   moduleDirectories: ['node_modules', '<rootDir>/'],
+  moduleNameMapper: {
```
mpinter, line 15: this deserves an explaining comment above and perhpas a mention in the docs as well

next/.env
```diff
@@ -18,3 +18,5 @@ NEXT_PUBLIC_IS_STAGING=true
 MSAL_CLIENT_ID=a6f86f77-53f7-4a8c-a1e8-06671a2395b8
 MSAL_SCOPE=https://graph.microsoft.com/.default
 MSAL_TENANT_ID=fe69e74e-1e66-4fcb-99c5-58e4a2d2a063
+
+OTP_SECRET=eFa02yTBjS96
```
mpinter, line 22: I guess these are forgotten - please remove

mpinter (changes requested):
just the couple small comments, but the approach is good :+1: thanks - let's just fix the couple eslint/ts issues before merging

will create a followup documentation task for this though - the requirement to write to call the validate function as async and the way ajvKeywords should be used should be written down somewhere

mpinter (approved):
✅

PR #448, opened 2022-12-09
Fix react-aria warnings

ValeriiMedvezhonkov:


mpinter (changes requested):
I don't think random uuid is a good candidate for label - looking at example usage in [react aria's docs](https://react-spectrum.adobe.com/react-aria/useRadioGroup.html#example), it seems more like we'd want to use the same text we use for - well, label. 

Sidenote - we also may want to display the label text in a label html component if we already don't do so.

next/components/forms/simple-components/ProgressBar.tsx
```diff
@@ -3,6 +3,7 @@ import { useProgressBar } from 'react-aria'
 
 type ProgressBarBase = {
   type?: 'success' | 'default'
+  label: string
```
mpinter, line 6: good intention, but I'm not sure if progress bar should always have a label - i.e. while it's not so in newest design, it used to be used as par of the header without any extra text

in the rest of the (widget) components the label makes more sense

PR #455, opened 2022-12-12
FORM-233: Error handling RJSF

AntonZezul:


AntonZezul:
Typescript strict mode errors, because validateRequiredFormat() dependence from useFormStepper which is not have typed vars. So validateRequiredFormat() need refactor when useFormStepper will refactored

next/backend/dtos/minio/upload-minio-file.dto.ts
```diff
@@ -1,6 +1,6 @@
 export interface UploadMinioFile {
   file: File
   originalName: string
-  errorMessage?: string
+  errorMessage?: string[]
```
matejpeluha, line 4: @AntonZezul This is not part of RJSF. This is error of file validation on client and also server. Please roll back to simple string and test upload if its still working like it should.

AntonZezul: Okey, my mistake

mpinter (approved):
Let's merge this, we'll be revisiting errors very soon. One thing I'd like to ask before this gets into master, as it should be easy enough:

- we'll need to stop displaying the list of errors on top of the form (as in the screenshot below) - we may replace the with our custom styled alerts in the future, but for now let's just get rid of them

Screenshot 2022-12-28 at 13 59 20

Things to note for the future:

- the error messages will need to be translated - the errors coming from our validations will need to be keys into the translation function instead of being the messages themselves
- if we have customFormats (and we'll likely have many more), we'll need to assign them custom error messages as well

PR #460, opened 2022-12-14
Ba 743

Balros:


radoslavzeman (approved):
Better than nothing :D 

Cards in org. structure have different width, depending on content.

image
image

PR #465, opened 2022-12-19
Form-188/summary

matejpeluha:
This is a combination of both tasks together FORM-188 and FORM-210 because it was necessary for me to test it in RJSF implementation at first for good code architecture. Of course, I added examples also to a style guide.


next/components/forms/steps/FinalStep.tsx
```diff
@@ -35,25 +51,25 @@ export const FinalStep = ({ state, slug }: FinalStepProps) => {
...
-        <div>
-          <pre>
-            {key}: {JSON.stringify(state[key], null, '\t')}
-          </pre>
-        </div>
-      ))}
-      {successMessage && <p>{successMessage}</p>}
-      {!!errors?.length &&
-        errors.map((error) => <p className="text-error">{JSON.stringify(error)}</p>)}
+      <h1 className="text-h1">Summary</h1>
```
mpinter, line 60: I understand this is to be fixed with the translations fix

mpinter (approved):
:+1: looking good - I think we will be revisiting the error handling in general in near future (still some questions about this on the product side) so I can't vouch for the errors part not needing changes, but imo we're good to merge 🚢

PR #466, opened 2022-12-19
FORM-208/rjsf-textarea-fix

matejpeluha:
Try to fix it but cant reproduce bug. This bug fix is just my intuition that it can help.
Mary used Windows 10 and Chrome. I am using Linux Mint and Chrome, Edge, Mozilla and Mindori but everything is ok at my system. If this fix would not work, it would be good, if somebody who is coding on windows would try to fix it.


mpinter:
I'm also not on windows machine so yes, let's try this out and see

PR #469, opened 2022-12-20
Form-184/Fix: Label warning

ValeriiMedvezhonkov:
Fix: Label warning, and custom width of progress bar was decreased because of on small displays it went beyond


mpinter (changes requested):
This means if I don't supply label I will get an ugly id written into a span above the progress bar - that's probably not intended.

ValeriiMedvezhonkov:
> This means if I don't supply label I will get an ugly id written into a span above the progress bar - that's probably not intended.

No, we render label ONLY if we put it as a prop, otherwise we need to add some labelId to useProgressBar hook. we don't get any ugly id on top of progress bar 
image

mpinter:
> > This means if I don't supply label I will get an ugly id written into a span above the progress bar - that's probably not intended.
> 
> No, we render label ONLY if we put it as a prop, otherwise we need to add some labelId to useProgressBar hook. we don't get any ugly id on top of progress bar, or you can just take a look on a styleguide, how it looks when progress bar doesn't contain any label image

Ah, sorry, misread the change, thought the `label: label ?? uuidv4(),` is in params not in `useProgressBar` hook.

All good then, LGTM

mpinter (approved):

PR #477, opened 2022-12-28
DEPRECATED Form 240/Field Groups RJSF

ValeriiMedvezhonkov:
Doubled input field RJSF widget and Input field + upload field are tested and seems it works correctly, I will rewrite and made code more clean


mpinter:
Same as the other DEPRECATED PR, closing and if we need to revisit will reopen.

PR #479, opened 2022-12-29
Form-168/Radio-fix-styling

ValeriiMedvezhonkov:
This component was already in "Done", but @AntonZezul  found this small styling issue, so I fixed component. However we should probably do more styling tests for all our components


next/backend/forms/test/uiSchema.json
```diff
@@ -158,7 +158,8 @@
       "ui:widget": "RadioButton",
       "ui:label": false,
       "ui:options": {
-        "className": "flex flex-col gap-4",
+        "className": "flex flex-row gap-4 mt-4",
+        "//className": "flex flex-col gap-4 mt-4 max-w-[500px]",
```
mpinter, line 162: is this on purpose or an error ? I didn't find this kind of syntax elsewhere

ValeriiMedvezhonkov: if you mean line 162 `"//className": "flex flex-col gap-4 mt-4 max-w-[500px]",` it is not an error, its just comment in json file how to use styling  if we want to use radio button in column order instead of row

mpinter: Ok, get it, wasn't familiar with commenting this way. Lgtm.

mpinter (approved):
🚢

PR #482, opened 2023-01-06
Form 240/RJSF Field Groups

ValeriiMedvezhonkov:


ValeriiMedvezhonkov:
@mpinter I don't see your comments here after our last call, if I forgot something to fix, let me know, otherwise all good ( still have issue to fix props problem like [FirstInputLabel - SecondInputLabel]. I don't know how to fix or use it in a proper way

mpinter:
@ValeriiMedvezhonkov apologies, left all the comments as 'pending' and forgot to submit the review. Anyway, most if not all are now outdated, will recheck the PR and let you know. I can also take a look at those types.

mpinter (changes requested):
Alright, a lot to dissect here, what I propose to do (check this with Mary as well and we're good to go down this way):

1) Let's omit the following fields completely:
```
    inputUpload,
    textareaUpload,
    inputSelect,
```

These I'm not confident with merging the way they are implemented today, they would require additional work and would stall this effort further at a time we need to move forward elsewhere. We may revisit later, but for now I'd just leave these components open in this branch and re-scope the original task to contain only the following:

```
    dateFromTo,
    timeFromTo,
    dateTime,
    doubledInput,
```

2) I've merged the proposed changes to working with types and added a bunch more comments - please follow up the type changes in the same manner on the rest of the components, and address the rest of the PR comments

3) this related to the first one so it's perhaps not a point on it's own :) but since we'll only want to merge the aforementioned 4 components, we want a PR with just these changes and nothing more. Will leave it up to you if you prefer editing this one or opening another with just those changes and the changes requested here.

Once all of the above is done we should be pretty close to merging this 🤞

ValeriiMedvezhonkov:
Pls, do not close o remove this PR or branch, we will use it for other (inputUpload, textareaUpload, inputSelect) components in a future

mpinter:
@ValeriiMedvezhonkov will close this just so it does not appear in open PRs, the branch will be kept and the PR can be reopened in the future

PR #493, opened 2023-01-11
Form 183/stepper

matejpeluha:
This is big branch for stepper + summary which is in progress.


next/components/forms/simple-components/Button.tsx
```diff
@@ -218,7 +218,7 @@ const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, PolymorphicProp
         className={style}
         {...buttonProps}
       >
-        <div className="flex items-center">
+        <div className="justify-center flex items-center">
```
mpinter, line 221: If this is to fix the issue you can see below on "Pokracovat" button, it probably does not work (at least not across all browsers, screenshot from Safari):

Screenshot 2023-02-10 at 10 02 56

next/components/forms/steps/Summary/Summary.tsx
```diff
@@ -0,0 +1,105 @@
...
+        (fieldFormData && !Array.isArray(fieldFormData)) ||
+        (Array.isArray(fieldFormData) && fieldFormData.length > 0)
+          ? JSON.stringify(fieldFormData, null, '\t').replaceAll('"', '')
+          : '-',
+      schemaPath,
+      isError,
+    }
+  }
+
+  const getAllSchemaData = (
```
mpinter, line 42: we can do this in a followup, but would like this function to be tested in isolation on a couple examples of prefilled form data, with and without errors

what that implies is for the function to live outside of the component, perhaps in one of the util files, which can be done as part of this PR

next/components/ui/FormRJSF/GeneratedFormRJSF.tsx
```diff
@@ -0,0 +1,97 @@
...
+  formSlug: string
+}
+
+const GeneratedFormRJSF = ({ eform, escapedSlug, formSlug }: FormRJSF) => {
+  const [t] = useTranslation('forms')
+  const form = useFormStepper(escapedSlug, eform.schema)
+
+  return (
+    <SectionContainer className="pt-14 md:pt-18">
+      {form.isComplete ? (
```
mpinter, line 26: afaik the stepper should be showing on the summary as well - I know designs have iterated here so if you've seen it without previously it was likely an error on our side (we apologise) https://www.figma.com/file/FHBgHIHB62YJjPLZcFMQxZ/ESBS---work-in-progress-(Mestské-konto%2C-Maťa)?node-id=560%3A71674&t=FVbZ8smVd12fMAhx-0

the summary should also be visible in the stepper, and "navigable" through stepper (you should be able to jump to summary and see what you've filled in so far)

next/pages/form/[eform].tsx
```diff
@@ -60,6 +63,20 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
   }
 }
 
+const initDefaultSchemaFields = (schema: StrictRJSFSchema) => {
```
Comment (mpinter) [inline next/pages/form/[eform].tsx:66]: would rather for such functions to live within utils files if they are outside of the component - though I see this is presently unused so if it's temporal until we decide what to do with it ok to keep here (but mark it with TODO please in such case, so that it's easiliy visible for anyone doing a cleanup of TODOs on the project)

next/utils/forms.tsx
```diff
@@ -0,0 +1,415 @@
+import Form from '@rjsf/core'
```
mpinter, line 1: why the .tsx rename ? I couldn't find any jsx within the file, you should be able to keep it as forms.ts as before (would prefer doing that)

I've skipped over this file for now as it's quite hard to track the changes within github ui with renamed files

mpinter (changes requested):
The stepper behaviour when jumping over steps is weird (1st click on whichever step which is not next step -> jumps to next step -> subsequent clicks on the same step does nothing -> click on any other step works as expected -> going to steps before that works as expected -> back to step 1) - see video:

https://user-images.githubusercontent.com/2327312/218050938-96aac009-112d-43a3-96fc-0890087ab444.mov

---

But it's starting to look good :+1: check the couple comments, but I'll be happy to merge this even before every detail is addressed as we'll iterate this further anyway.

next/components/forms/simple-components/Button.tsx
matejpeluha: @mpinter do you mean missing icon in Pokracovat button ? i ddnt know until now that there s bug like that, but i would say its not problem on my side. Its problem on button component and I do not think i should fix thing slike these that with amount of work on this task.

next/utils/forms.tsx
matejpeluha: i forgot to rename it back to ts. I was just trying one trick few few weeks ago. Thanks !

next/utils/forms.ts
```diff
@@ -364,16 +357,19 @@ export const useFormStepper = (eformSlug: string, schema: RJSFSchema) => {
     increaseStepErrors()
     setStepFormData(newFormData)
     const isFormValid = await validate()
+    console.log(isFormValid)
```
mpinter, line 360: forgotten log

next/components/forms/simple-components/Button.tsx
mpinter: I see, did not realise it's becasue of the icon, likely a black icon on black background, though it's bad padding. Yes, as you've said, that's outside of scope of this task.

mpinter (approved):
:+1:

Alright, let's merge this bad boy 🚢

PR #495, opened 2023-01-11
Form-183/stepper-view-desktop

matejpeluha:


next/components/forms/steps/StepperView.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface StepperViewProps {
+  steps: StepData[]
+  currentStep: number
+  onChangeStep?: (stepIndex: number) => void
+}
+
+const StepperView = ({ steps, currentStep, onChangeStep }: StepperViewProps) => {
+  return (
+    <div className="p-3">
+      {steps.map((step: StepData, key: number) => (
```
mpinter, line 13: key is defined as optional string in TransformedFormData and as number here - the way order is used below it probably won't work nicely when undefined or arbitrary string is passed in

mpinter (approved):
Check out the comment below before merging, otherwise good to go.

next/components/forms/steps/StepperView.tsx
matejpeluha: That is different property key which has nothing to do with key from map() function. But I chose wrong naming convention because as I see, its confusing. Thanks!

PR #497, opened 2023-01-12
FORM-251 login

kristianduda:
Need to confirm mobile design with Patrik. Think, that code can be checked right now.


mpinter (approved):
:+1: , few renames & lgtm

PR #507, opened 2023-01-16
FORM-256 forgotten password

kristianduda:


next/components/ui/AccountSuccessAlert/AccountSuccessAlert.tsx
```diff
@@ -0,0 +1,33 @@
+import ArrowRightIcon from '@assets/images/forms/arrow-right.svg'
```
mpinter, line 1: would like to separate these sections from the UI of bratislava.sk - this is mainly because much of what presently resides within components/ui will need refactoring and I want to separate newly created components (which use slightly different design) from this cohort

we're putting all the components using this 'new' design into `forms` - this directory might get a rename to be more fitting (let's do that elsewhere), as it will contain also sections from city account, but for now let's place everything there

while on this topic - please find a reasonable subdirectory for this within `/forms`, perhaps `/forms/segments` ? would call it a `section`, but section already has a connotation within our CMS :) .. naming things is hard

you can do this in a followup PR and do it with the 'sections' (or components) from the login PR as well if there are any that currently resides within ui, I likely missed this in that PR

mpinter (approved):
:+1: good to go, but a couple of followups:

1. It might be a security feature on the side of Cognito - but is there presently a way to detect that password reset was successfully initialised, as opposed to incorrect email being entered ? Current behaviour is that no matter if you input existing or non-existent email, you will always be sent to the second step. If there is please fix and don't send user to the second screen if their email is not within system. If there isn't let me know and we'll work on the wording on the second screen.

2. There is this visual issue I might have missed previously, with the header overlapping the content on mobile:

https://user-images.githubusercontent.com/2327312/212661957-b5998a76-962c-421b-bd58-a068b4819002.mov

3. check the comment regarding component hierarchy

PR #508, opened 2023-01-16
Form 249/RJSF markdown in accordion

ValeriiMedvezhonkov:
@mpinter Is it the right way I am using markdown content in UI schema for fields?

This branch an PR only for testing


mpinter:
@ValeriiMedvezhonkov yup :+1: , I think this makes sense, let's do it

ValeriiMedvezhonkov:
@mpinter Waiting for approve Field-Group RJSF task, before finishing this task, pls review existing code

next/components/forms/widget-wrappers/WidgetWrapper.tsx
```diff
@@ -30,6 +36,26 @@ const WidgetWrapper = ({
       })}
     >
       {children}
+      {Array.isArray(accordion) &&
+        accordion.map((item) => {
+          return (
+            <Accordion
+              key={uuidv4()}
```
mpinter, line 43: instead of this use this approach (version for mapped elements) - https://bratislava.github.io/frontend/components#use-generated-ids

next/utils/utils.ts
```diff
@@ -52,3 +52,6 @@ const isServer = () => typeof window === 'undefined'
 export const isBrowser = () => !isServer()
 
 export const isProductionDeployment = () => process.env.NEXT_PUBLIC_IS_STAGING !== 'true'
+
+export const isObject = (value: any) =>
```
mpinter, line 56: I assume it's from here ? https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript if so (and in fact, even if it isn't) please add this link for context

```suggestion
// https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript
export const isObject = (value: any) =>
```

mpinter (approved):
Please fix the two comments before merging but otherwise good to go

PR #513, opened 2023-01-17
Form-240/RJSF-FieldGroups

ValeriiMedvezhonkov:
dateFromTo,timeFromTo,dateTime,doubledInput field group widgets


next/components/forms/widget-components/DateTimePicker/TimeSelector.tsx
```diff
@@ -69,9 +76,13 @@ const TimeSelector = ({
 
   useDidMount(() => {
     const hoursItemOffset =
-      hoursItemRef && hoursItemRef.current ? hoursItemRef.current[+hour]?.offsetTop : 0
+      hoursItemRef && hoursItemRef.current
+        ? hoursItemRef.current[+timeFormatArray[0]]?.offsetTop
+        : 0
```
mpinter, line 81: missed this on last review - but again, this is much nicer with ?.


```suggestion
      hoursItemRef?.current?.[+timeFormatArray[0]]?.offsetTop || 0
```

next/components/forms/widget-components/DateTimePicker/TimeSelector.tsx
```diff
@@ -69,9 +76,13 @@ const TimeSelector = ({
...
     const minutesItemOffset =
-      minutesItemRef && minutesItemRef.current ? minutesItemRef?.current[+minute]?.offsetTop : 0
+      minutesItemRef && minutesItemRef.current
+        ? minutesItemRef?.current[+timeFormatArray[1]]?.offsetTop
+        : 0
```
mpinter, line 85:
```suggestion
      minutesItemRef?.current?.[+timeFormatArray[1]]?.offsetTop || 0
```

next/components/forms/widget-components/DateTimePicker/TimeSelector.tsx
```diff
@@ -124,7 +138,13 @@ const TimeSelector = ({
                     clickHandler(e, 'minute', `${item}`)
                   }}
                   className={cx('cursor-pointer rounded-lg px-10 py-2 focus:outline-none', {
-                    'bg-gray-100': +minute === item,
+                    'bg-gray-100': +timeFormatArray[1] === item,
+                    'cursor-pointer pointer-events-auto opacity-100':
+                      (minValueArray && minValueArray[0] < timeFormatArray[0]) ||
```
mpinter, line 143:
```suggestion
                      (minValueArray?.[0] < timeFormatArray[0]) ||
```

elsewhere as well

mpinter (changes requested):
Please test thoroughly - there are multiple issues with filling in those fields, resubmit when you are certain that they behave as they should.

ValeriiMedvezhonkov:
@mpinter 
this PR includes much more work than it should(need to have separate tasks, but ok), was tested all date and time pickers [also in field group with From-To logic]. Also I reviewed and found UI bug with field aligning and error messages direct in components so I decided to write custom logic to display erorr messages in Field-Group components. ( to save tester time in a future)

- * One bug still ( actually not a bug a small mistake in time format) exist in time picker. If I type a time in a field ( not via popup) for example in a H:MM for example "4:15" ( correct time format will be "04:15") format and go next (**_Question:_** correct time format is HH:MM ???? ) in a formData object it will be in H:MM format instead of HH:MM, but after revisiting page (Go previous) time format will be changed in a correct way, I tried to do some hack like 'componentWillUnmount' in a functional component but it didn't work, so I left as it is.

- * For Date and Time field-group widgets was changed logic and now we MUST use predefined keys such as [startDate-endDate or startTime-endTime]. Maybe we can deal in this way in other components ( not only for this 4 component, I mean InputSelectWidget and so on). If so, we can solve the problem with indexas for fields.

next/pages/form/[eform].tsx
```diff
@@ -94,15 +93,59 @@ const FormTestPage = ({
       })
     })
   }
+  const validateDateFromToFormat = (formData: any, errors: FormValidation) => {
+    const formDataKeys = Object.keys(formData)
+    formDataKeys?.forEach((key) => {
+      if (
+        form?.currentSchema?.properties[key]?.dateFromTo &&
```
Comment (mpinter) [inline next/pages/form/[eform].tsx:100]: this assumes this data lives only on 'top level' of each step, which might not be true. But same issue is in the other validation, I'm ok with fixing this all at once and keeping like this for now, just pointing this out.

mpinter (changes requested):
Can't get to 3rd step, will be shown this error:

Screenshot 2023-01-27 at 11 41 31

When I click on 'next' and the validation does not pass, the data in whole step will be reset to their initial value, this is not wanted.

ValeriiMedvezhonkov:
> 

maybe you forgot to pull last changes, or try to remove node_monudeles and install once again ( idk where is problem cause I tested it ant its works fine.

mpinter (approved):
alright, pulled again, restarted the server, all works

let's 🚢 (finally! 🙂)

PR #514, opened 2023-01-17
FORM-203: Orientation prop and fix after review

AntonZezul:


AntonZezul:
@mpinter 
I used brackets (min-w-[24px]), because tailwind doesn't have min-w for this size that we need

mpinter (approved):
:+1: looks reasonable, 🚢

PR #516, opened 2023-01-18
Form-178/Fix toggle button

ValeriiMedvezhonkov:
made controlled state logic, fix css styles

### DIFF
--- a/next/components/forms/simple-components/Toggle.tsx
+++ b/next/components/forms/simple-components/Toggle.tsx
@@ -12,8 +12,9 @@ type ToggleBase = {
   isReadOnly?: boolean
   defaultSelected?: boolean
   isSelected?: boolean
-  children: React.ReactNode
-  value: string
+  children?: React.ReactNode
+  value?: string
+  onChange?: (isSelected: boolean) => void
 }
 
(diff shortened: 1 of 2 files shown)


matejpeluha (approved):
It looks like problems are fixed. Thanks

mpinter (changes requested):
We've been recently dealing with how to do components "correctly" across our projects, for this component in particular the part about delaing with controlled state is important: https://bratislava.github.io/frontend/components#use-controlled-state

This is something we did not have in place when beginning this project, but we should now move towards those principles. That means that the logic of how the state is handled should look very close to (for example) this component from Marianum project https://github.com/bratislava/marianum/blob/master/next/components/atoms/TagToggle.tsx

---

On other note - `value` should probably be renamed to `id` and dealt with this way: https://bratislava.github.io/frontend/components#use-generated-ids

matejpeluha:
@mpinter Its not a critical part, I do not need it right now. I have used Toggle in user profile where its needed, but its just not working correctly. It would be good, if this is done next monday/tuesday, but its ok if it will take longer

ValeriiMedvezhonkov:
@mpinter Issue with value-id was fixed, controlled state was already did in the way you proposed. Please re-review once again

PR #518, opened 2023-01-19
FORM-233: Error handling after review

AntonZezul:


mpinter (approved):
Did not check the async change, but will trust it now works as expected :)

PR #528, opened 2023-01-23
FORM-250 register

kristianduda:


mpinter (approved):
Honestly did not test this enough, just a single happy-path pass through, but I don't want to hold this further.

Please before merging:
- take a look at the errors mentioned below
- take a look at the visual issue with the buttons in the following screenshot - the lower button does not have correct height (visible only
 on hover) and both the texts are not centered. If this is a problem with the underlying component let me know and I'll create a followup bug issue, if it's easy to fix here please fix right away.

Screenshot 2023-01-24 at 13 56 19

PR #531, opened 2023-01-24
BA-670: Adjust quote markdown design

JacobMu:
[JIRA task](https://inovaciebratislava.atlassian.net/browse/BA-670)

image


next/components/atoms/HomepageMarkdown.tsx
```diff
@@ -144,7 +144,7 @@ export const HomepageMarkdown = ({ className, content, numericalList }: Homepage
             <div className="flex gap-x-8 lg:gap-x-6">
               <div
                 className={cx(
-                  'h-4 w-4 shrink-0 bg-category-600 rounded-full mt-1 border-4 border-solid border-category-600',
+                  'h-4 w-4 shrink-0 bg-category-600 rounded-full mt-2 border-4 border-solid border-category-600',
```
mpinter, line 147: should this change be here ? does not seem related, and if it is it also changes other lists which is not wanted

mpinter (commented):
check / remove the change for `li` component (or let me know fi I misunderstood), otherwise looks fine

next/components/atoms/HomepageMarkdown.tsx
JacobMu: This change was to adjust a bullet point more to center with the first line. But now when I look into it, the previous value was good enough.

mpinter (approved):

PR #544, opened 2023-01-30
Form-230/Tile component

ValeriiMedvezhonkov:


next/components/forms/simple-components/ServiceCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+  target,
+  className,
+}: ServiceCardBase) => {
+  const style = cx(
+    'w-[280px] bg-gray-0 border-gray-200 flex flex-col items-start p-5 gap-5 border-solid border-2 rounded-lg',
+    className,
+  )
+
+  return (
+    <a href={href} target={target} rel="noreferrer">
```
mpinter, line 31: this is a visual component , it shouldn't contain the linking behaviour - often you'll be wrapping this in Next's Link https://nextjs.org/docs/api-reference/next/link because you'll be linking within the same application. Please remove.

next/components/forms/simple-components/ServiceCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <div role="button" tabIndex={0} className={style} aria-hidden="true">
```
mpinter, line 32: why would you do `aria-hidden="true"` ? this will hide the whole subcomponent from the screenreaders. Also why `tabIndex={0}` ? a static value does not make sense here, you do not know where these cards will be used.

next/components/forms/simple-components/ServiceCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <div className="min-w-[75px] min-h-[75px] max-w-[75px] max-h-[75px] flex items-center justify-center">
```
mpinter, line 33: once again - you are using a lot of classes which are not really needed, I think you can get pretty much the same behaviour here with just:

```suggestion
        <div className="w-[75px] h-[75px]">
```

next/components/forms/simple-components/ServiceCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <Logo />
+        </div>
+        <div className="gap-3 flex flex-col items-start w-full">
+          <div className="text-h-base font-semibold not-italic">{title}</div>
```
mpinter, line 37: Why `not-italic` (also elsewhere) ? There is no italic definition to override. Please go through all of this and remove classes which are doing nothing.

next/components/forms/simple-components/ServiceCard.tsx
mpinter: also I don't think you need role="button", that will be handled by the wrapping <Link> or <a> (as mentioned in the other comment, outside of the visual component) - was this just forgotten ?

mpinter (changes requested):
Please clean-up the css, also take a look at [this slack thread](https://inovaciebratislava.slack.com/archives/C0458KY6YG1/p1675162216105919) (key part was copied into the JIRA task as well) and update the hover interaction and behaviour within rows accordingly.

next/components/forms/simple-components/ServiceCard.tsx
```diff
@@ -0,0 +1,50 @@
...
+  href,
+  className,
+}: ServiceCardBase) => {
+  const style = cx(
+    'group min-w-[280px] max-w-[280px] bg-gray-0 border-gray-200 flex flex-col items-start p-5 gap-5 border-solid border-2 rounded-lg cursor-pointer',
+    className,
+  )
+
+  return (
+    <Link href={href}>
```
mpinter, line 30: we still did not understand each other - even the Link should not be part of the component itself. Just unwrap it from this, we'll be adding either Link or `<a>` tag from the parent component later when we'll be using the cards

mpinter (approved):
One more point talked about in the [linked slack thread](https://inovaciebratislava.slack.com/archives/C0458KY6YG1/p1675162216105919) and in the [JIRA task](https://inovaciebratislava.atlassian.net/browse/FORM-230) wasn't addressed - the icon/image should be with a transparent background, the background color should be chosen as another param.

Apart from this and the Link behaviour mentioned below it should be good - pre-approving not to hold this back, please fix these two before merging and we're good to go.

PR #545, opened 2023-01-31
FORM-261 zmena korešpondenčnej adresy

kristianduda:


next/components/forms/segments/CorrespondenceAddressModal/CorrespondenceAddressModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+  show: boolean
+  onClose: () => void
+}
+
+const CorrespondenceAddressModal = ({ show, onClose }: Props) => {
+  const { t } = useTranslation('forms')
+  const [errorMessage, setErrorMessage] = useState('')
+
+  const onSubmit = ({ data }: { data?: CorrespondenceAddressData }) => {
+    console.log(data)
```
mpinter, line 18: I think you can update in cognito right as a part of this - or is there something blocking such updates ?

or since this is now more on the level of just being a visual component, maybe just move the onSubmit to the parameter and supply the cognito onSubmit wherever we'll actually use this ?

mpinter (approved):
Did not run this but code looks fine 🙂 lgtm.

PR #562, opened 2023-02-06
FORM-262: Add waiting and error cards

AntonZezul:


mpinter (changes requested):
Please ask for english version of texts from product and place the text into translation files.

With longer paragraphs like these, it might be worth it having it saved as markdown within the translation file and have it rendered as markdown within the component.

mpinter (approved):

PR #567, opened 2023-02-07
FORM-234 header

kristianduda:


next/components/forms/segments/AccountNavBar/AccountNavBar.tsx
```diff
@@ -86,15 +125,86 @@ export const AccountNavBar = ({ className, currentLanguage }: IProps) => {
               </p>
             }
           />
+          <nav className="text-font/75 flex gap-x-8 font-semibold">
+            <div className="text-font/75 flex items-center gap-x-6 font-semibold">
+              {!navHidden ? (
+                <>
+                  <Link href="/" variant="plain">
```
mpinter, line 132: please verify accessibility guidelines (I don't have a good reference unfortunately, but search around the web for some baseline for webpage menus) - would at least expect a list element somewhere around here, maybe some aria markers as needed ?

mpinter: in fact, a good baseline might be to use https://react-spectrum.adobe.com/react-aria/useMenu.html - we're trying to move to react-aria everywhere it makes sense. The useMenu seems to handle these kinds of menus well, should take care of accessibility and should let you style as needed.

Same for other menus in this PR.

next/components/forms/segments/AccountNavBar/AccountNavBar.tsx
```diff
@@ -118,14 +228,182 @@ export const AccountNavBar = ({ className, currentLanguage }: IProps) => {
...
+    <div className="relative flex w-12 cursor-pointer items-center" ref={ref} onClick={handleClick}>
+      <div className="font-light lg:font-semibold">{current?.toUpperCase()} </div>
+      <ChevronDownSmall
+        className={`ml-3 hidden mix-blend-normal lg:flex ${
+          isSelectClicked && isComponentVisible && 'mb-1 -rotate-180'
+        }`}
+      />
+      {isSelectClicked && isComponentVisible && (
+        <div className="absolute top-6 -left-3 z-20 mt-1 flex h-auto w-11 cursor-default flex-col items-center justify-center lg:left-0">
+          <div className="z-10 h-0 w-4 border-x-8 border-b-4 border-solid border-transparent border-b-[#F8D7D4]" />
```
mpinter, line 391: we shouldn't have much need for custom colors (like border-b-[#F8D7D4]) - if the color is part of our desing (system) it should also be part of the tailwind.config (or our colors config in css file, don't know where they presently live)

next/components/forms/segments/AccountNavBar/AccountNavBar.tsx
```diff
@@ -118,14 +228,182 @@ export const AccountNavBar = ({ className, currentLanguage }: IProps) => {
...
+          <div className="flex h-auto min-h-[60px] w-full flex-col items-center rounded-lg bg-[#F8D7D4] pt-1 pb-3 shadow-[0_8px_24px_rgba(0,0,0,0.16)]">
```
mpinter, line 392: bg-[customcolor] should be replaced. I also think we should have all the required boxShadow configsin tailwind.config - you should be able to use one of them, or nudge the design if this one is not one of the preconfigured ones to go along with the ones in design system. Same elsewhere.

mpinter (changes requested):
Please check below, but otherwise looking good.

next/components/forms/segments/AccountNavBar/AccountNavBar.tsx
```diff
@@ -11,10 +10,12 @@ import useAccount, { UserData } from '@utils/useAccount'
 import cx from 'classnames'
 import HamburgerMenu from 'components/forms/segments/HambergerMenu/HamburgerMenu'
 import Button from 'components/forms/simple-components/Button'
+import Menu from 'components/forms/widget-components/Menu/Menu'
```
mpinter, line 13: probably not widget-components but .. simple-components ?

the reasoning behind "widget" components is that they should all be used as RJSF widgets

we'll likely be doing a cleanup of components structure when we start merging bratislava.sk more closely with with account page

mpinter (approved):
:+1:, check the couple renames & lgtm

PR #571, opened 2023-02-07
FORM-275: Update Tooltip

AntonZezul:


mpinter (changes requested):
This is making the scope a bit larger, but it's fixing an issue I did not know we had - we shouldn't be doing the tooltip trigger on our own, instead we should use https://react-spectrum.adobe.com/react-aria/useTooltipTrigger.html . There are multiple a11y considerations built into it which we currently don't respect.

Please rewrite it to use the aforementioned hook. Using the hook, you can still control the opened/closed state and can do so to satisfy the requirements of the FORM-275 task.

mpinter (commented):
check the comment in slack, otherwise looking good

mpinter (approved):
:+1: 🚢

PR #584, opened 2023-02-14
FORM-229: Fix rjsf error in DoubleInputField

AntonZezul:


AntonZezul:
We have a problem with schema.json, default prop doesn't working, maybe need to wait for Stepper component and if-then-else logic

PR #588, opened 2023-02-14
Form-259/TaxHeaderContactInfo_Visual

ValeriiMedvezhonkov:
Created page for taxes with header and contact info, added modal window for changing correspondence address


next/components/forms/segments/AccountSectionHeader/TaxFeeSectionHeader.tsx
```diff
@@ -0,0 +1,140 @@
...
+      return statusNode(<SuccessIcon width={18} height={24} viewBox="0 0 24 18" />, 'Uhradená')
+
+    default:
+      return null
+  }
+}
+
+const priceFormat = (price: number): string => {
+  return Number.isInteger(price) ? `${price},00 €` : `${price} €`.replace('.', ',')
+}
```
mpinter, line 49: would rather not have assumptions about the format this comes from the server - please remove this fn and in place where you use it, just keep a placeholder text

next/components/forms/segments/AccountSectionHeader/TaxFeeSectionHeader.tsx
```diff
@@ -0,0 +1,140 @@
...
+      </div>
+      <div className="flex flex-col items-start gap-2 w-full h-full">
+        <div className="text-p2-semibold text-main-700 sm:block hidden">{t('city')} Bratislava</div>
+        <div className="flex flex-col items-start gap-4 h-full w-full">
+          <div className="flex flex-row items-center gap-4 w-full">
+            {/* Desktop Title */}
+            <div className="text-h1 grow sm:block hidden">Daň z nehnuteľností</div>
+
+            {/* Mobile Title */}
+            <div className="text-h1 grow sm:hidden block">Daň z nehnuteľností 2023</div>
```
mpinter, line 76: waiting for answer here https://inovaciebratislava.slack.com/archives/C0458KY6YG1/p1676391722195559, if the answer is that the title should be the same for mobile and desktop then pelase change, if not good as is

PR #590, opened 2023-02-15
Form-266/PaymentData_Footer_Visual

ValeriiMedvezhonkov:
Added Payment tax data page, created desktop and mobile version.
Created tax footer.
Added translations for payment data part


next/components/forms/segments/AccountSections/TaxesFeesSection/TaxFooter.tsx
```diff
@@ -0,0 +1,13 @@
+import parse from 'html-react-parser'
```
mpinter, line 1: please don't use this

if you need a (somehow) structured text in translation use markdown

if markdown is not enough for what you need, it's ok, just split the translation part into more segments (you can use sub-objects within the translation parts to group related translations together, check out https://github.com/bratislava/bratislava.sk/pull/620/files#diff-06f870440cf0ccbec161a218c290ba04b013aaffd8d69ac65a51b9184caad4a5R31 ), and keep the html/css/react part in code outside of translations

mpinter (changes requested):
let's get rid of the react-parse lib, otherwise just minor changes

next/components/forms/simple-components/ClipboardCopy.tsx
```diff
@@ -0,0 +1,20 @@
...
+const ClipboardCopy = ({ copyText }: { copyText: string }) => {
+  const [_, copy] = useCopyToClipboard()
+  return (
+    <button
+      type="button"
+      onClick={() => {
+        copy(copyText)
+          .then(() => {})
+          .catch(() => {})
+      }}
```
mpinter, line 13: we did not understand each other the first time - I meant this

```suggestion
      onClick={() => copy(copyText)}
```

mpinter (approved):
Please see the suggestion (would much prefer it to land in master that way) - the rest looks good.

PR #596, opened 2023-02-17
FORM-285 fix rc tooltip, validation

kristianduda:


next/components/forms/info-components/Tooltip/TooltipPopup.tsx
```diff
@@ -51,7 +51,9 @@ const TooltipPopup = ({ arrow = true, className, text, position = 'top-left' }:
           className,
         )}
       >
-        <p className="w-max">{text}</p>
+        <p className="w-max">
+          <ReactMarkdown>{text || ''}</ReactMarkdown>
```
mpinter, line 55: Please use "AccountMarkdown" (merged just today), it should ensure that we have correct typography weights (and helps us with fixing them across every markdown usage should they change)

mpinter (approved):
✅

PR #608, opened 2023-02-20
FORM-213: Design system - breakpoints

AntonZezul:


mpinter (changes requested):
I still see a few `2lg` classes, these shoul,d probably be cponverted to lg

Other than that fine :+1:, just waiting for product feedback.

mpinter (approved):
Ok, we're good to go, lgtm

PR #614, opened 2023-02-21
FORM-285 tooltip

kristianduda:
fix rc and id card tooltip, change message on identity verification, get access token on login


next/pages/register.tsx
```diff
@@ -92,12 +95,23 @@ const RegisterPage = ({ page }: AsyncServerProps<typeof getServerSideProps>) =>
             </AccountSuccessAlert>
           )}
           {status === AccountStatus.IdentityVerificationRequired && (
-            <IdentityVerificationForm onSubmit={verifyIdentity} error={error} />
+            <IdentityVerificationForm
+              onSubmit={(rc, idCard) => {
+                // bold values
+                setLastRc(`**${rc}**`)
+                setLastIdCard(`**${idCard}**`)
```
mpinter, line 102: the bold markings should be in the account.json, not in this value

next/utils/useAccount.ts
```diff
@@ -347,6 +347,7 @@ export default function useAccount(initStatus = AccountStatus.Idle) {
           })
           AWS.config.credentials = awsCredentials
 
+          accessToken = result.getAccessToken().getJwtToken()
```
mpinter, line 350: it's strange how accessToken is used here - or rather, it's not at all clear. 

It seems the only time it's used is in verify identity, that function is expecting it will have the correct instance coming from the closure, but that's relying on either the login function or the useEffect being called in the 'same render' .. ? Is this correct or am I missing something ?

If the above is the case, wouldn't it be preferred to rewrite accessToken to be a state variable ? `const [accessToken, setAccessToken] = useState('')`

mpinter (changes requested):
Might be out of scope of the original task, but would like to clean-up (or get explained) the accessToken usage before moving this forward.

mpinter (approved):

PR #616, opened 2023-02-21
FORM-289: Fix Aktualne Section after review

AntonZezul:


ValeriiMedvezhonkov:
Safari icons problem was i .svg files, we need to have strict width and height, not '100%'. exists 2 ways of solution: change direct in svg file width and height from '100%' to xxPX or provide as a prop to svg component width and height.
Screenshot 2023-02-27 at 11 35 57

mpinter (approved):
@ValeriiMedvezhonkov the solution where we provide the size to the svg component as a prop sounds good

✅

ValeriiMedvezhonkov (approved):

PR #617, opened 2023-02-21
Form 257/my profile

matejpeluha:


next/components/forms/segments/AccountMarkdown/AccountMarkdown.tsx
```diff
@@ -52,7 +58,10 @@ const AccountMarkdown = ({ className, content, variant = 'normal' }: AccountMark
           a: ({ href, children }: { href?: string; children?: string }) => (
             <UILink
               href={href ?? '#'}
-              className="break-words font-semibold text-font underline hover:text-category-600"
+              className={cx('break-words font-semibold  underline hover:text-category-600', {
+                'text-white': uLinkVariant === 'primary',
+                'text-font': uLinkVariant === 'default',
+              })}
```
mpinter, line 64: I'd say this is again one of those where it's inconsistent in the design and it should get cleaned up, instead of adding variants to the code

next/components/ui/UserProfile/UserProfileView.tsx
```diff
@@ -56,16 +57,18 @@ const UserProfileView = () => {
       updateUserData(temporaryUserData).then(() => {
         setIsEditing(false)
         setIsAlertOpened(true)
+        setTimeout(() => setIsAlertOpened(false), 3000)
```
mpinter, line 60: would prefer to have all of `handleOnSubmitEditing` as async instead of having part of it in `then` callback. You also want to handle the catch part - show an error when the cognito write fails (and useful info if you can determine why it failed)

mpinter (approved):
Merging to move this forward before holidays but we'll need to revisit, both for the comments (which are minor) and for the phone validation / header (which is inconsisten with rest of the pages). I'd also prefer having this under `/account` url.

PR #618, opened 2023-02-23
FORM-298: Mam problem Section

AntonZezul:


AntonZezul:
Banner was no finished, i will create separate PR to solve all Banners on accont page

PR #623, opened 2023-02-23
Form-323/StepperMobileAnimation

ValeriiMedvezhonkov:
slide animation from bottom to top


mpinter (changes requested):
Go to summary screen -> click on 'edit' next to one of the fields -> when you're moved to the given step the fullscreen stepper slides in and needs to be closed (it's probably related to these changes as it's not appearing, even without animation, on master when these steps are reproduced).

But other than that looks good :+1:

mpinter (approved):

PR #653, opened 2023-03-01
FORM-335 account provider

kristianduda:


next/utils/useAccount.tsx
```diff
@@ -70,10 +70,42 @@ export interface AccountError {
   code: string
 }
 
-export default function useAccount(initStatus = AccountStatus.Idle) {
+interface Account {
```
mpinter, line 73: prefer type over interface (no specific reason for this case, just trying to move to a unified approach within the project)

mpinter (approved):

PR #655, opened 2023-03-01
Form-324/Modal_phoneNumber_functionality

ValeriiMedvezhonkov:
update phone_number to user profile


mpinter (approved):
Please fix the couple comments beforehand

Also - this will require a followup for the task to be complete - we only want to show this once. WIll write in slack regarding this.

PR #656, opened 2023-03-02
reset rc

kristianduda:
Possibility to reset birthnumber. Also enabled tier attribute update in Cognito (for testing purposes).


next/pages/get-jwt.tsx
```diff
@@ -10,17 +11,36 @@ import { Wrapper } from 'components/styleguide/Wrapper'
...
+      }
+    } catch (error) {
+      console.log(error)
+    }
   })
 
+  const resetRc = async () => {
+    try {
+      await resetRcApi(accessToken)
+      const res = await updateUserData({ tier: null })
```
mpinter, line 36: whoa, this can be done ? this resets the data in cognito or just locally ?

would expect this not to be possible, and the approach here being reloading the user data from cognito after resetRcApi

mpinter (approved):

next/pages/get-jwt.tsx
kristianduda: I've set write permission to tier attribute in Cognito console. Just for the test. 
Will not be possible in production.

PR #659, opened 2023-03-02
FORM: Update Banner component on account page

AntonZezul:


AntonZezul:
I changed Button component - add a new prop fullwidth because i couldn't change w-full style in className

mpinter (approved):

PR #683, opened 2023-03-06
Form 210/rjsf summary fix

matejpeluha:


next/components/forms/steps/Summary/StepButtonGroup.tsx
```diff
@@ -0,0 +1,44 @@
...
+  isFinalStep: boolean
+  previous: () => void
+  skip: () => void
+  submitStep: () => void
+  submitForm: () => void
+}
+
+const StepButtonGroup = (props: StepButtonGroupProps) => {
+  const { stepIndex, isFinalStep, previous, skip, submitStep, submitForm } = props
+  const [t] = useTranslation('forms')
```
mpinter, line 18: it probably does not matter, but we use the `{}` syntax  (`const { t } = useTranslation('forms')` ) everywhere else, so let's do it here as well

mpinter (approved):
did not test, looks ok, 🚢

PR #685, opened 2023-03-06
Sections cleanup

MarekBodingerBA:
Commit by commit.

All sections from Sections.tsx are extracted do their own components, files related to them moved to the same directory, fixed types for UI components that were receiving wrong types, clean up of GraphQL queries.


next/graphql/queries/General.graphql
```diff
@@ -0,0 +1,63 @@
+query Footer($locale: I18NLocaleCode!) {
+  footer(locale: $locale) {
+    ...Footer
+  }
+}
+
+fragment Footer on FooterEntityResponse {
```
radoslavzeman, line 7: Consider if `FooterFragment` won't be misinterpreted as fragment `on Footer` while it is on `FooterEntityResponse`. We used to do fragments as `Footer on Footer` and `FooterEntity on FooterEntity`.

radoslavzeman (approved):

PR #698, opened 2023-03-07
Official board

MarekBodingerBA:


mpinter (approved):
Did not run this but the code looks fine :+1: , wanted to pitch in just to reconfirm the api should indeed not be touched yet as it is changing as we speak - the plan is to bring in a lib in which we'll maintain our GINIS integrations (so yes, current approach looking good).

next/components/molecules/sections/general/OfficialBoardSection/OfficialBoardSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+    queryFn: () =>
+      fetch(
+        `/api/ginis/documents-list?${new URLSearchParams({
+          search,
+        })}`,
+      ).then((r) => r.json() as Promise<ParsedOfficialBoardDocument[]>),
+    keepPreviousData: true,
+  })
+
+  return (
```
radoslavzeman, line 24: Where was this code before? I don't see any code like this removed in this PR so maybe you left some unused code somewhere?

MarekBodingerBA:
> Did not run this but the code looks fine 👍 , wanted to pitch in just to reconfirm the api should indeed not be touched yet as it is changing as we speak - the plan is to bring in a lib in which we'll maintain our GINIS integrations (so yes, current approach looking good).

So, I should revert the changes on the API and create a new one or is it fine just like it is now?

next/components/molecules/sections/general/OfficialBoardSection/OfficialBoardSection.tsx
MarekBodingerBA: From `next/pages/mesto-bratislava/transparentne-mesto/uradna-tabula.tsx`, I will migrate the pages first and then remove the old code in a new PR.

mpinter:
> > Did not run this but the code looks fine 👍 , wanted to pitch in just to reconfirm the api should indeed not be touched yet as it is changing as we speak - the plan is to bring in a lib in which we'll maintain our GINIS integrations (so yes, current approach looking good).
> 
> So, I should revert the changes on the API and create a new one or is it fine just like it is now?

It's fine as it is - I understand tis as not really touching the api 🙂 - the code parsing the response is the same, it was just moved

PR #722, opened 2023-03-09
FORM-297: Municipal Services - fix filter and pagination

AntonZezul:


AntonZezul:
Fixed filter and pagination, select field and dropdowns solving in another PR.

mpinter:
Started reviewing, but would rather discuss on a call before continuing

PR #733, opened 2023-03-13
FORM-332: fit menu items width to bold content

kristianduda:


mpinter (approved):
Did not test, but seems reasonable 🙂

PR #739, opened 2023-03-13
Form-264/TaxDetails_MapTestData

ValeriiMedvezhonkov:
map test data to accordion, fix modal height for large content


ValeriiMedvezhonkov:
All data in AccordionTableContent.tsx file only for test purposes, but data formats are real, from BE

mpinter (approved):
✅

PR #942, opened 2023-05-13
Fix index

MarekBodingerBA:


MarekBodingerBA:
https://github.com/bratislava/bratislava.sk/assets/104828482/e5eed70b-ae47-48e4-b127-3889b762a1be

The reason why this is happening.

PR #1084, opened 2023-09-29
Bump react-aria-components

radoslavzeman:
- update to use new beta version
- update Tabs according to new docs
- install new `tailwindcss-react-aria-components` plugin
- install recommended `tailwindcss-animate` so we can test it slowly
- fix rgb syntax - in modern css, rgb values are written without commas - this also fixes usage with opacity (e.g. `bg-gray-800/50`)


radoslavzeman:
Fixed. Also added script to bump all related react-aria libraries. (copied from konto.bratislava.sk project)

PR #1100, opened 2023-10-25
updated node18

skifahrer:
Updated strapi/next to support node 18.18.2. Already on staging, we have this version for testing with copied prod db.


next/Dockerfile
```diff
@@ -1,11 +1,14 @@
...
 
 FROM base AS build-base
 WORKDIR /build
 COPY ./package.json ./yarn.lock ./
 
 FROM build-base AS install-prod
 ENV YARN_CACHE_FOLDER=/build/.cache/yarn
+# https://forum.strapi.io/t/docker-strapi-node-18-16-alpine-sharp-node-gyp-problem-solved/32245
+# https://stackoverflow.com/questions/71966803/error-sh-node-gyp-build-command-not-found
+RUN yarn global add node-gyp
```
vidriduch, line 11: Ok, so couple of things here. Not a big fan of adding this as a separate dependency. Based on the links shared, there is a problem with `node 18.18 alpine` image, where the issue is. 

Also not a big fan of keeping the version so strict, e.g., `node:18.18.2-alpine` for our team setup. As basically you are committing yourself to upgrade all images at least once a month. Because, when you freeze version like this, if they release any security patch you will not get it automatically. But if you keep it just `node:18-alpine` you will.
The other side of this argument is, that unintentionally upgrading base image can break the app, which is fair point but IMHO in our current team setup we can afford occasionally fixing image version, but can't afford to spent time once a month to check for all security release and upgrading all image versions.

With that said as I see it we have two options:
1. We can keep `yarn global add node-gyp` and change the version to `node:18-alpine`, which should work, _based on the discussion in the links you provided_ (didn't test it). Then we will get security upgrades and this should still work. Then make a note of it (or set a reminder or something) and when they fix it, we remove it.

2. Or we can remove this dependency `yarn global add node-gyp` and fix the version on `node:17-alpine` as mentioned in these comments 

> downgrading to node:18.17-alpine seems to work for me 18.17.1 is a security release ...
> It is probably better to use node:18.17-alpine

and then again make a note of it and try to fix it latter, when another version of a node image will be released (or we upgrade node to higher version).

If we choose to keep the dependency, there is something to be said about adding this as a separate layer (separate `RUN` statement), where I'm torn between having two layers and shorter built time but potentially slightly bigger image and having this in included in the following `RUN` statement, but smaller image. In current setup both are "sort of" fine, especially if we don't keep this dependency there for long, so your pick.

vidriduch (changes requested):
Overall this is fine, but some discussion needs to be had about how strict we want to be with node image versions and adding new dependencies

next/Dockerfile
vidriduch: Well yes, that is the other side of the argument that I was talking about. And yes you probably want to keep this as strict as possible. But again, this happen once in two years? If you keep the version stuck to the bugfix version (18.18.**x**), you will not even get those and the image version will need to be updated or check for updates, once a month? Which means if there is a change in version and bugfixes came you will have to go and iterate through all of the projects and their images to change it and rebuilt.

We are already doing this for IaC stack and it takes it's tool. My only concern is, that in current setup we don't have the man power to keep this upgraded. But If you feel like we can afford this (which is fine as keeping strict version is and should be a standard) I'm fine with keeping the version strictly fixed, but then I will hold you up on the promise of updating it once in a while 😄

vidriduch (approved):
From my side LGTM. Just check with @radoslavzeman or @mpinter if you want to add those `.nvmrc` files. I'm not sure if we have those also for other projects and if we want to commit them. 
If not let's add them into `.gitignore` if they are fine, we can merge.

PR #1122, opened 2023-11-24
exec aria upgrade script + fixed error on input after upgrade

ValkovMirec:
- update of react aria packages
- resolving error on ref in input components


daksas303:
Removed the necessity for the user to pass their own ref from outside even when not needed. But we are still keeping the option to pass it, this might be useful for cases like when using `react-hook-form`.

Issue #1135, opened 2023-12-13
Remove `// @ts-strict-ignore`


radoslavzeman:
Continuously remove all `@ts-strict-ignore` comments.

Firstly, always check if the file or functions in the file are used. If not, remove the whole file.

Secondly, try to fix errors that appear when removeing `// @ts-strict-ignore`. Do not use `any`, do not disable eslint rules, but try to fix the code. Use any or disable the rule only if no other option is possible.
Do not spent too much time on the file or issue, ask quickly if you get stuck.

Read more in readme: https://github.com/bratislava/bratislava.sk/blob/f3d5a1b86bec15cad63e2738e62598fcee3c37b9/next/README.md#typescript-strict-mode

PR #1142, opened 2023-12-15
Remove ts-strict-ignore from components and fix issues

daksas303:
- https://github.com/bratislava/bratislava.sk/issues/1135
- Removed `@ts-strict-ignore` from components
- Deleted `ArticlesFilter` as it doesn't seem to be used. Considering the TODO there, I am unsure if it was kept as a base for some potential redesign.


radoslavzeman:
Most changes were copied from here and merge in #1145

PR #1151, opened 2023-12-20
Fix continuously eslint warnings and errors

daksas303:
- Continuously fixing eslint problems where possible

### DIFF
--- a/next/components/forms/Alert.tsx
+++ b/next/components/forms/Alert.tsx
@@ -38,74 +38,72 @@ const Alert = ({
   message,
   className,
   ...rest
-}: AlertBase) =>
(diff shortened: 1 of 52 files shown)


radoslavzeman (commented):
In `.map` function, a unique `key` must be provided.

If we have a unique prop, that can be used, please us it (usually `id` or `slug`).
If we don't have a unique prop, then we use `index` and we disable the eslint rule locally. It's safer to use index that non-unique value in this case.

Use name `index` instead of `i`, we don't use abbreviations.

Please, continue with other files than `forms` directory. I would go through all other components than `forms` and `ui`, then `ui`, and then finally `forms`.

PR #1182, opened 2024-02-24
Cypress init

sebastian-vass:
#1181
- Inicializácia cypress, eslint, prittier.
- Nastavenieconfigov pre tento projekt.
- Pridaná dokumentácia na spúštanie testov.


karolinavor (approved):

tests/cypress.config.ts
```diff
@@ -0,0 +1,31 @@
...
+      config.video = false
+      config.baseUrl = config.env.BASEURL
+      
+      config.env = {
+        devices: {
+          desktop: ['all', 'desktop'].includes(config.env.DEVICE),
+          mobile: ['all', 'mobile'].includes(config.env.DEVICE),
+        },
+        resolution: {
+          desktop: { viewportWidth: 1440, viewportHeight: 1080 },
```
radoslavzeman, line 18: Where do these numbers come from?

tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+Currently tests are running against http://localhost:3000 (If you need to change the baseURL you need to do it in the ./cypress.sh), so the app needs to be running. Either follow the e2e setup in [next app readme](../next/README.md), or rebuild & restart via the following npm script from this directory:
+
+```
+yarn start:e2e
+```
+
+## Run Cypress locally
+
+```
+npm run cypress:open //for open mode (Running tests in chrome browser)
```
radoslavzeman, line 12: Do we use `yarn` or `npm`?

tests/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "lib": ["es5", "dom", "es2016.array.include"],
+    "types": ["cypress", "cypress-file-upload", "node"],
+    "allowJs": true,
+    "isolatedModules": false,
+    "baseUrl": "./",
+    "noEmit": true,
+    "skipLibCheck": true
+  },
+  "include": ["**/*.ts"],
+  "exclude": ["node_modules"]
```
radoslavzeman, line 13: Does this setup corresponds with other projects? (not only Konto)

tests/.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-env node */
```
radoslavzeman, line 1: What is this comment for?

tests/tsconfig.json
radoslavzeman, line 1: I'm missing path alias to be able to import by full path. (something like `"@/*": ["./*"]` and basepath)

tests/cypress.config.ts
sebastian-vass: For the width I used the value that is set in the dev tools under Laptop L. I left the height at 1080px, through experience. It is better to use smaller values than larger ones.

tests/README.md
sebastian-vass: The use of both is not exclusive, but I will rewrite it so that only yarn will be used everywhere.

sebastian-vass: fixed

tests/tsconfig.json
sebastian-vass: Currently I've push the basic tsconfig, which is needed to avoid errors. The original tsconfig contained things for the future that were needed in the Konto. I assumed that the full tech stack on both projects is 1:1. That's why I hadn't considered that it wouldn't be necessary.

tests/tsconfig.json
sebastian-vass: I have added support and fixtures that will be used most often.

tests/.eslintrc.js
sebastian-vass: As I write [above](https://github.com/bratislava/bratislava.sk/pull/1182#discussion_r1504610642).

tests/cypress/support/e2e.ts
```diff
@@ -0,0 +1,10 @@
+import './commands/global'
+import './triggers/before'
```
radoslavzeman, line 2: I'd love to see absolute imports here, today's common practice is to use `@/...whatever` (see the comment in tsconfig)

tests/tsconfig.json
radoslavzeman: I would prefer to set
```
    "paths": {
      "@/*": ["./*"]
    }
```
so it covers all imports and does not have to be maintained. Then we will be able to use `import ... from @/cypress/support/...` etc.

The prefix `@/` instead of just `@` is used to distinguish between npm packages and local full paths.

I know we have it set up differently in most projects but this seems to be newest best practice. See `paas-mpa` repo for example.

tests/cypress/support/commands/global.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-namespace */
```
radoslavzeman, line 2: Please, never disable eslint rules for the whole file.
When needed, always disable it locally for specific line or codeblock.

You can put it around codeblock like:
```
/* eslint-disable some/rule */
   some codeblock here, for example a function or interface...
/* eslint-enable some/rule */

radoslavzeman (commented):
Few more comments or questions, please discuss with @mpinter before implementing.
Otherwise it looks cleaner now :)

mpinter (approved):
Please check the comments above - change the eslint disable comments to block scoped and, as agreed on call, try out the path alias as suggested and see if it works out of the box - if it doesn't please comment in this PR and leave as it is.

Either way you can merge without further review.

PR #1194, opened 2024-03-01
Nocnapomoc.sk redirection

skifahrer:
As request from IT department, I added redirection from nocnapomoc.sk to our bratislava.sk subpage


mpinter (approved):
well, it looks exactly like the other ingress setups, which worked, so probably good ? :)

PR #1241, opened 2024-03-27
change codegen .yml to .ts

radoslavzeman:


mpinter (approved):
makes sense , including the comment (would also prefer doing it against schema, but ok to do against a server)

Issue #1319, opened 2024-07-30
Create `ContactsSearchSection`


radoslavzeman:
Create section based on `<GlobalSearchSectionContent variant="specific" searchOption="users" />`, similar to `RegulationsListSection` or `OfficialBoardSection`

- aktualnost so Silviou?

PR #1327, opened 2024-08-21
#1319 Add Contacts search section

radoslavzeman:
TODO:
- create Strapi section
- hide pagination - MS Graph does not have pagination, only page size (`$top`) that specifies what is the max count of returned results
- fix bug when we show "No results found" during first loading


radoslavzeman:
Not planned.

PR #1328, opened 2024-08-23
Added livenessProbe & PDB. Change resource limits

vidriduch:
* Change `readinessProbe` to run more often with deployment becoming unready after 30sec
* Added `livenessProbe` to restart the deployment after 90sec
* Added PDB to increase availability in production environment
* Change resource requests & limits based on long term monitoring


radoslavzeman (approved):
👍

PR #1350, opened 2024-09-23
Persist dockerized postgres data locally

radoslavzeman:


docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
       - '5432'
     ports:
       - '127.0.0.1:5432:5432'
+    volumes:
+      - ./postgres-data:/var/lib/postgresql/data
```
vidriduch, line 15: (optional) I would rather keep it consistent with meilisearch. As meili is mounting into `meilisearch` folder I would rather have this be mounting into `postgresql` folder to match the mount path. So either 

```suggestion
      - ./postgresql:/var/lib/postgresql/data
```

or 

```suggestion
      - ./postgresql:/var/lib/postgresql
```

vidriduch (approved):

docker-compose.yml
radoslavzeman: using `./postgresql:/var/lib/postgresql`

PR #1369, opened 2024-11-28
Node maintenance

skifahrer:
Upgraded node versions and dependencies where possible.


strapi/package.json
```diff
@@ -36,13 +36,13 @@
...
   "engines": {
-    "node": ">=18.x.x",
+    "node": "18.x.x",
     "npm": ">=9.8.1",
     "yarn": "1.22"
   },
   "license": "MIT",
   "volta": {
-    "node": "18.17.1",
+    "node": "18.18.0",
```
radoslavzeman, line 45: In next/Dockerfile, you use v20, here I see v18. Shouldn't it be the same?

PR #1402, opened 2025-02-25
Fix cypress test for search results

radoslavzeman:


jankaBehal (approved):
thank you, lets try this

PR #1404, opened 2025-02-26
Replace local ginis wrapper with ginis-sdk

lukasbabula:
Node was updated to version "20.9.0" to fulfill requirements of the ginis-sdk library. To achieve this, minio needed to be updated to version "^7.1.3". Since this version, it no longer requires "@types/minio" as it is now included automatically, so it was removed. Please review this minio change very carefully.


next/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bratislava-next",
-  "version": "1.4.0",
+  "version": "1.5.0",
```
radoslavzeman, line 3: No need to change this in FE projects. :)

radoslavzeman (approved):
Nice! Looks very promising 👐

next/package.json
lukasbabula: Ok, I will take note for any future changes

Issue #1446, opened 2025-03-25
Redirect `bratislava-pre-ukrajinu/asistencne-centrum-bottova`


radoslavzeman:
SK
from `bratislava-pre-ukrajinu/asistencne-centrum-bottova`
to `bratislava-pre-ukrajinu/asistencne-centrum-loom`

EN
from `en/bratislava-for-ukraine/assistance-centre-of-help-on-bottova-street`
to `en/bratislava-for-ukraine/assistance-centre-loom`

PR #1447, opened 2025-03-25
#1446 Add redirect for `asistencne-centrum-bottova`

radoslavzeman:


jankaBehal (approved):
ok thank you for redirect

PR #1451, opened 2025-03-26
#1450 Update eslint plugins

radoslavzeman:
Update according to OLO.


jankaBehal (approved):
Everything seems to work as expected, thank you

PR #1455, opened 2025-03-28
Validate ginis ids

lukasbabula:
also change Ginis endpoint to HTTP


radoslavzeman (approved):
Nice job! Thank you 🤍

PR #1460, opened 2025-03-31
#1459 Use eslint plugins next, i18n, tanstack

radoslavzeman:


jankaBehal (approved):
thank you for this change

Issue #1398, opened 2025-02-21
Make the menu non-sticky


radoslavzeman:
The sticky menu occupies to much space (even more significant it is on smaller laptop screens).

Make the menu non-sticky and allow it to scroll out. (Similar as in olo project)

PR #1471, opened 2025-04-07
#1398 make the menu non sticky

jankaBehal:
- ### Add line instead of shadow
image

- ### Remove stickness

image

- ### Remove hidden empty block
- ### Add scroll to top button

image


radoslavzeman (commented):
Menu is now over the "decoration":

image

radoslavzeman (approved):

Issue #1453, opened 2025-03-26
Implement GTM


jankaBehal:
Implementácia Google Tag Manager pre potreby nasadenia cookie consentu a Google Analytics
According to the OLO implementation, for container id refer to @filipranik 
https://github.com/orgs/bratislava/projects/49/views/11?pane=issue&itemId=98578611&issue=bratislava%7Colo.sk%7C1104

PR #1476, opened 2025-04-11
#1453 Implement GTM

jankaBehal:
Add Google Tag Manager to envs - Prod, Staging, empty for Dev, because it is not set in GTM to work with dev
Remove Google Analytics from envs
Rename plausibleProps -> analyticsProps

Notes: some parts might be missing from config-sync, which is currently out of sync, it will be added in next PR


next/package.json
```diff
@@ -33,6 +33,7 @@
     "@bratislava/ginis-sdk": "^1.1.0",
     "@internationalized/date": "3.7.0",
     "@js-joda/core": "5.6.4",
+    "@next/third-parties": "^15.3.0",
```
radoslavzeman, line 36: Please use correct version (same as `next`).

next/.env.development
radoslavzeman, line 1: Note for future us: We want to change this file to `.env.example` (not in this PR)

radoslavzeman (approved):
Thank you! 🤍

PR #1484, opened 2025-04-16
1481 bug inba page does not react to links refresh is needed

radoslavzeman:


jankaBehal (approved):
Thank you for this change

PR #1545, opened 2025-05-23
#1414 Announce current page in Pagination

katarinazaprazna:
### Description

Improve pagination accessibility

Changes:
- Add new translation strings to differentiate between the current page ("Page 3, current page") and navigable pages ("Go to page 3")
- Use the existing `selected` prop to set `aria-current`, creating consistency with the visual styling logic
- Remove the unused `aria-current` prop from the pagination item type returned by usePagination hook

These changes ensure screen reader users can clearly identify which page they're currently viewing

### Screenshots

Screenshot 2025-05-23 at 14 54 23


radoslavzeman (changes requested):
Hmm this should work as in Material design without custom handling.

PR #2098, opened 2026-02-11
1810 job page list of open positions

BardunovaBarbora:
- added new section jobOfferList to strapi
- added new nextjs component JobOffersSection with list of (for now) mocked data that are to redirect user to the details of the job offer (in Nalgoo)
- added api call to Nalgoo
- added client pagination
 **TODO: add secret for Nalgoo api to .env/.env.develop**

image

### DIFF
--- a/next/src/components/cards/JobOfferRowCard.tsx
+++ b/next/src/components/cards/JobOfferRowCard.tsx
@@ -1,5 +1,5 @@
 import { Typography } from '@bratislava/component-library'
-import React, { ElementType } from 'react'
+import React from 'react'
 
 import { ChevronRightIcon } from '@/src/assets/icons'
 import CardBase from '@/src/components/cards/CardBase'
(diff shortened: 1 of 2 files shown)


next/src/components/sections/JobOffersSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Figma: https://www.figma.com/design/A9aoQH2FGhR1D14wvvk6FW/Mestsk%C3%BD-web--bratislava.sk-?node-id=3331-7668&t=42jF8eIC6pPGhocy-4
+ */
+
+type JobOffersSectionProps = { section: JobOfferListSectionFragment }
+
+const JobOffersSection = ({ section }: JobOffersSectionProps) => {
+  const { t } = useTranslation()
+  const { title, text, titleLevel = 'h2' } = section
+  const titleId = useId()
```
radoslavzeman: The `titleId` is used to connect card title and link with `aria-labelledby`. It needs to be unique for each card. Now, when the card is not separate component, all cards get the same titleId, so all cards will have same title read by screen reader.

The easy solution is to move the card to separate component. :)  (That will also be consistent with other cards)

next/src/components/sections/JobOffersSection.tsx
```diff
@@ -0,0 +1,124 @@
...
+    salary: '1500 EUR/mes',
+    salaryInfo: 'od 1500 EUR/mes.  v závislosti od pracovných schopností, praxe a skúseností ',
+    employmentForms: [{ id: 1, name: 'Plný úväzok' }],
+    fields: [],
+  },
+]
+
+/**
+ * Figma: https://www.figma.com/design/A9aoQH2FGhR1D14wvvk6FW/Mestsk%C3%BD-web--bratislava.sk-?node-id=3331-7668&t=42jF8eIC6pPGhocy-4
+ */
```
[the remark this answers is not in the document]
radoslavzeman: Nice, that was quick! :)

next/src/components/cards/JobOfferRowCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+              className="group-hover:underline"
+            >
+              {title}
+            </Typography>
+          </MLink>
+          <Typography variant="p-tiny">
+            {[location, employmentForms, salary, salaryInfo].filter(isDefined).join(' • ')}
+          </Typography>
+        </div>
+        <ChevronRightIcon className="self-center max-lg:invisible" />
```
radoslavzeman: I think that the reason to not show icon on mobile is to safe some space.
Class `invisible` hides the icon, but keep the space for it. Therefore, if my assumption is correct, we should use `max-lg:hidden` instead of `max-lg:invisible`.

PR #2102, opened 2026-02-12
Update style of LinkRowCard component

BardunovaBarbora:
- truncated link to 1 row
- added aria-hidden
- plain button

image


radoslavzeman (approved):
Please add also figma link :)

Issue #2066, opened 2026-01-28
Uradna tabula - Add Year field


radoslavzeman:
Add `Year` field to additional filters on "Uradna tabula" search.

Additional logic:
- Option "All years" should be disabled or hidden when "archived" documents are selected. It's due to large number of documents  and long requests (querying whole archive can take more than 10 seconds). Therefore we split querying archive by years.
- Default value for "published" documents should be "All years". The reason is, that in january, we cannot show only current (new) year, because there will be also some published documents from previous year that must still be visible.

Note: Keep it hidden on production, until fully merged and tested on staging.

Figma (just for inspiration, not full design):
https://www.figma.com/design/A9aoQH2FGhR1D14wvvk6FW/Mestsk%C3%BD-web--bratislava.sk-?node-id=1225-4767&t=LNxZUjaq2Y8UdHKe-0

<img width="2432" height="400" alt="Image" src="https://github.com/user-attachments/assets/c174bf62-3b47-4257-b768-41d9c3d95843" />

PR #2106, opened 2026-02-16
#2066 Add new year field to official board

BardunovaBarbora:
- Added new publishedYear filter to official board
- Removed number of documents in categories

image

### DIFF
--- a/next/src/components/sections/SearchSection/SearchResults.tsx
+++ b/next/src/components/sections/SearchSection/SearchResults.tsx
@@ -35,7 +35,7 @@ const SearchResults = ({
 
   const searchQuery = useQueryBySearchOption({ optionKey: searchOption.id, filters })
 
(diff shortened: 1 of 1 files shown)


next/src/components/sections/SearchSection/SearchResults.tsx
```diff
@@ -48,7 +48,11 @@ const SearchResults = ({
     onSetResultsCount(searchOption.id, searchResultsCount ?? 0)
   }, [onSetResultsCount, searchOption.id, searchResultsCount])
 
-  if (isPending) {
+  useEffect(() => {
+    onLoadingChange(isRefetching)
+  }, [isRefetching])
```
radoslavzeman: Hmm let's discuss this since it relies on Tanstack logic and it affects the main search page.

PR #2130, opened 2026-02-25
#2113 #2135 Create new Disclosure component

BardunovaBarbora:
- Created new Disclosure comonent 
- added to Accordion section
- added to faq section
- added to footer
- removed old Accordion component
TODO: do we use FaqCategoriesSection? if not, because now i see we use AccordionSection in faq page, do we remove it?

image

image

### DIFF
--- a/next/src/components/sections/AccordionSection.tsx
+++ b/next/src/components/sections/AccordionSection.tsx
@@ -38,7 +38,7 @@ const AccordionSection = ({ section }: AccordionSectionProps) => {
       {/* TODO Correct spacing between SectionHeader and remaining content */}
       <div className="flex flex-col gap-6 lg:gap-8">
         <SectionHeader title={title} titleLevel={titleLevel} />
-        {flatText!.length ? (
+        {flatText?.length ? (
           <DisclosureGroup
             allowsMultipleExpanded
             className="rounded-xl border border-border-active-default bg-background-passive-base py-2"


next/src/components/common/Disclosure/index.ts
radoslavzeman: Please, do not create `index.ts` barrel files. In the ES module system, it adds an unnecessary step and behaviourally leads to creating fewer components. 
For consistency, please removi it and import components directly from their files.

next/src/components/common/FaqsGroup/FaqsGroup.tsx
```diff
@@ -15,9 +21,18 @@ const FaqsGroup = ({ faqs, accordionTitleLevel }: FaqsGroupProps) => {
   return (
     <div className="flex flex-col gap-4">
       {faqs.filter(isDefined).map((faq, index) => (
-        <Accordion key={index} title={faq.title} accordionTitleLevel={accordionTitleLevel}>
-          <Markdown content={faq.body} variant="small" />
-        </Accordion>
+        <DisclosureGroup allowsMultipleExpanded key={index}>
+          <Disclosure id={`disclosure-faq-${index}`}>
```
radoslavzeman: Please use `faq.id` / `faq.documentId`.

Good practice: Allways use database ids when possible. Only fallback to index if there is no database id.

radoslavzeman (commented):
It looks as a very good progress!

I'm thinking if we can make it more strict and parametrised. Now, it feels too loose (header text size, dividers...). But we can discuss this in person.

next/src/components/common/Disclosure/index.ts
BardunovaBarbora: fixed

next/src/components/common/Footer/MobileFooter.tsx
```diff
@@ -37,17 +41,26 @@ const MobileFooter = () => {
           <FooterContacts {...footer} />
         </div>
         <div>
+          <DisclosureGroup className="border-0">
```
radoslavzeman: To discuss

next/src/components/common/Disclosure/DisclosurePanel.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <AriaDisclosurePanel
+      {...props}
+      className={cn(
+        'px-4 lg:px-6',
+        // animation
+        'h-(--disclosure-panel-height) overflow-clip motion-safe:transition-[height]',
+        props.className,
+      )}
+    >
+      <div className="pb-4">{children}</div>
```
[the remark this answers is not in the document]
radoslavzeman: To discuss

PR #2150, opened 2026-03-02
2135 new accordion

BardunovaBarbora:
added new disclosure instead of native details/summary tags in MobileBreadcrum and ArticlesFilters
Blocked by https://github.com/bratislava/bratislava.sk/pull/2130

image

### DIFF
--- a/next/src/components/sections/AccordionSection.tsx
+++ b/next/src/components/sections/AccordionSection.tsx
@@ -38,14 +38,14 @@ const AccordionSection = ({ section }: AccordionSectionProps) => {
       {/* TODO Correct spacing between SectionHeader and remaining content */}
       <div className="flex flex-col gap-6 lg:gap-8">
         <SectionHeader title={title} titleLevel={titleLevel} />
-        {flatText!.length > 0 && (
+        {flatText!.length ? (
           <DisclosureGroup
             allowsMultipleExpanded
             className="rounded-xl border border-border-active-default bg-background-passive-base py-2"
           >
             {flatText?.filter(isPresent).map((item, index) => (
               <Fragment key={`disclosure-${item.category}-${index}`}>
-                {index > 0 ? <HorizontalDivider aria-hidden className="mx-4 lg:mx-6" /> : null}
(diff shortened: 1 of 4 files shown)


next/src/components/sections/AccordionSection.tsx
```diff
@@ -17,54 +24,38 @@ type AccordionSectionProps = {
...
-                      .filter(Boolean)
-                      .filter(isDefined)}
-                    url={file.url ?? undefined}
-                    urlLabel={file.urlLabel ?? undefined}
-                  />
-                ))}
-              </div>
-            </Accordion>
-          ))}
+        {flatText!.length > 0 && (
```
radoslavzeman: Never use `non-null assertion`. (This eslint rule should be marked as error, not just warn)

Instead, consider using `flatText?.length ? (...) : null`. This is our standard approach.
Notice also the change to ternary operator - this is crucial to prevent rendering "0" for an empty array.

next/src/components/sections/AccordionSection.tsx
```diff
@@ -48,7 +48,7 @@ const AccordionSection = ({ section }: AccordionSectionProps) => {
                 {index > 0 ? <HorizontalDivider aria-hidden className="mx-4 lg:mx-6" /> : null}
```
radoslavzeman: I like that you are thinking about accessibility and you are actually using aria attributes!

In this case however, there is no need to use `aria-hidden`, because `<HorizontalDivider>` has role="reparator", that serves the same purpose. 

Please remove aria-hidden from here :)

next/src/components/common/FaqsGroup/FaqsGroup.tsx
```diff
@@ -14,10 +17,21 @@ export type FaqsGroupProps = {
 const FaqsGroup = ({ faqs, accordionTitleLevel }: FaqsGroupProps) => {
   return (
     <div className="flex flex-col gap-4">
```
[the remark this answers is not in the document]
BardunovaBarbora: will adress in future as agreed

PR #2151, opened 2026-03-02
#2138 Use new fields `metaDescription` and `path` in pages

BardunovaBarbora:
- changed `metaDiscription` to new `metaDescription` in Pages
- added `path` to Pages


radoslavzeman (approved):
Nice, thanks!

PR #2196, opened 2026-03-18
#1341 Fix bug with page path

BardunovaBarbora:


radoslavzeman (commented):
Please also rename PageBySlug to PageByPath.
Also check:
- getLinkProps
- useQueryBySearchOption
and other fulltext search may help.

Please also remove `slug` from graphql.

PR #2265, opened 2026-04-09
#2233 Implement official board filter preserved state

BardunovaBarbora:
- added categoryId, publicationState, publicationYear into one filter state
- remove forgotten documents meilisearch


radoslavzeman:
After the discussion with @BardunovaBarbora and @Ty-ci, we close this PR without merging.
We agreed that Official Board filtering will be moved to separate component, following same approach as in DocumentsAll section.

PR #2307, opened 2026-05-11
Revert #1872 - remove automatic page hierarchy

radoslavzeman:
- remove prototype of automatic page hierarchy introduced in #1872 because it wasn't used even after a few months and it clutered code and strapi
  - if we want to implement it in future, we can still reintroduce it


BardunovaBarbora (approved):
nice 👍

PR #2364, opened 2026-06-16
#2365 Fix job offers order and redundant spaces

### LINKED ISSUE #2365: JobOffers reversed order
Show job offers in reversed order to match nalgoo https://bratislava.nalgoo-jobs.com/

Trim values to get rid of redundant trailing spaces.

radoslavzeman:
- revers the order to match nalgoo page
- trim incoming values from nalgoo because they ofter contain redundant spaces
- fix `Error: Calling setState synchronously within an effect can trigger cascading renders` by deriving totalItems count from data instead of using effect


BardunovaBarbora (approved):
looks good

PR #2465, opened 2026-08-20
Bump strapi memory to 6Gi from 3Gi

mpinter:
Seems reasonable still ? And we seem to be ratelimiting the /graphql bots to at least some degree now. What do you think about bumping the RAM 2x ?


vidriduch (approved):
Sure, but can I get some details why is it that strapi needs more memory? I guess those are spikes? So, is it that someone is uploading large documents? Or is it something else?

mpinter:
@vidriduch I don't think so, this time it were actual page revalidates (so actual pulling of CMS data). But, this last instance of trouble was because of an error on web's side, which caused a lot more of the revalidates happening than usual. Other spikes are still suspected to be bot traffic. But - while both solvable by other means (and being worked on), I think this was still running on memory a little too close to the limit where such spike put it out of balance.

But - since these don't seem to be happening because of giant file uploads, I think the "proper" fix here is to finetune the GC to work better, so that it collects the allocated memory before it piles up to OOM. TODO

PR #2503, opened 2026-09-01
AVI-to-nginx-migration: upload limits and proxy header buffers

mpinter:
Note: should be merged once production default ingress is switched to nginx.

Part of the AVI → nginx cutover. Adds to the ingress:

```yaml
nginx.org/proxy-buffer-size: 16k
nginx.org/proxy-buffers: 8 16k
```

Responses on `*.bratislava.sk` carry large cookies, and nginx's 4k default response-header buffer rejects them with a 502 and `upstream sent too big header` in the controller log. That is exactly what happened to `city-account-next` on staging after its cutover (infra #350), and AVI had no such limit, so it only surfaces once traffic moves.

Both keys are required together — `nginx -t` rejects `proxy_buffer_size 16k` on its own, because the default `proxy_buffers 8 4k` then violates `proxy_busy_buffers_size`.

Verified with `kustomize build` on the Dev/Staging/Prod overlays: the only change in the rendered output is the two added annotations.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/next/kubernetes/base/ingress.yml
+++ b/next/kubernetes/base/ingress.yml
@@ -6,6 +6,8 @@ metadata:
   labels:
     service: app
   annotations:
+    nginx.org/proxy-buffer-size: 16k
+    nginx.org/proxy-buffers: 8 16k
     cert-manager.io/cluster-issuer: cloudflare-dns01
     cert-manager.io/issue-temporary-certificate: 'true'
 spec:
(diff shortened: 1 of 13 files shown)


mpinter:
Superseded: `proxy-buffer-size`/`proxy-buffers` now go into the nginx controller ConfigMap instead — bratislava/infrastructure-deployment-configuration#372. One entry covers every ingress on the cluster, including the ones this PR touched, so there is nothing left for this change to do.

